PR #9, opened 2015-08-13
MUMUP-2045 add description for MyUW logo - add NAME.ariaLabelTitle

jiayinjx:


timlevett:
:+1:  Friendly reminder you will have to do this in angularjs-portal [here](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/master/angularjs-portal-home/src/main/webapp/js/app-config.js) and [here](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/master/angularjs-portal-home/src/main/webapp/js/master-app-config.js)

jiayinjx:
Thanks!

PR #11, opened 2015-08-13
Remove less from frame

jhanstra:
Moving marketplace.less, home.less, widget.less, and uw-portlet.less to /web. Anything else to move?

Also, have I done everything to ensure things don't break? Still trying to get a grasp of how overlays work.


timlevett:
:+1: Pulled down, reviewed the frame pages (home, notifications, and settings. All looks good!)

PR #12, opened 2015-08-19
Add in circle button directive

timlevett:
We use circle buttons a lot, lets just make a directive. I give you `<circle-button>`.

Options:
+ href : the href you want the anchor to have
+ target : the target (if blank, it'll launch in same window)
+ fa-icon : the fa-icon you want to use (e.g. fa-star)

Here is a demo:

image

### DIFF
--- a/src/main/webapp/portal/misc/directives.js
+++ b/src/main/webapp/portal/misc/directives.js
@@ -170,6 +170,18 @@ define(['angular', 'require'], function(angular, require) {
             link: linker
         };
     });
+    
+    app.directive('circleButton', function() {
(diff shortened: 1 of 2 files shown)


vertein:
:+1: I like that the whole circle is a button

apetro:
:+1: :100: Excellent refactor.  Love it.  Exactly the kind of discovering and executing on opportunities for simplification and re-use that will keep us agile.

keirserrie:
:+1:

jhanstra:
late to the party, but :+1:

PR #13, opened 2015-08-20
MUMUP-2016: Features Page and Revamped Popup

jhanstra:
## Features Page
Adds a Features Page at /features that will display the contents of the features.json feed in a structured blog-like way, in order by id (in reverse). So when adding a feature, simply add an item to the array in features.json with the next id and fill in the attributes.

Example:
image
image

### Features of the Features Page (so meta)
- Features are shown according to features.json which is app-specific, so any uw-frame app makers can have their own list of features showcased
- A user-facing popup modal (like what we did for the welcome page on 8/11) can be enabled or disabled for new features. This is great in that it allows us to announce new, smaller features subtly in a way that doesn't interrupt the user. They are just added to the Features page for viewing.
- Popup will show the latest feature in features.json (if that feature is enabled for a popup - if not, it will not be shown) and will only be shown once. It still uses local storage to keep track of whether it's been seen.
- Prominent button at bottom of the Features page linking to Feedback page

### features.json
The structure looks like this:

```
{
   "id" : 1,
   "title" : "Welcome to the new MyUW",
	 "description" : "The new MyUW makes it easier to access your content. You can switch between simple icons and more informative widgets using 'Compact' and 'Expanded' toggles. Each of the icons on the home page links to your MyUW content, and widgets can display a dashboard of information on the home page. If you are having trouble finding what you're looking for, you can use the top search box to search through all of MyUW's content, and save it to your home page from there using the 'Add to home' button.",
   "img" : "img/myuw-welcome.png",
   "learnMoreURL" : "http://google.com",
	 "goLiveYear" : 2015,
	 "goLiveMonth" : 6,
	 "goLiveDay" : 11,
	 "popup" : {
		 "enabled" : false,
		 "startYear" : 2015,
		 "startMonth" : 7,
		 "startDay" : 11,
		 "endYear" : 2015,
		 "endMonth" : 7,
		 "endDay" : 25,
		 "buttonText" : "Continue to MyUW"
	 }
}
```

### Possible Areas for Further Improvement
- Keep track of whether the user has seen each popup-enabled feature, and if there is more than one they haven't seen, show them both in a single step-through-able popup.
- More customizable popups


src/main/webapp/css/buckyless/features.less
```diff
@@ -0,0 +1,46 @@
+.features-feed {
```
timlevett: :+1: :100: for this

src/main/webapp/portal/features/partials/features.html
```diff
@@ -0,0 +1,31 @@
...
+	        <h3>{{ feature.title }}</h3>
+	        <p>{{ feature.description }}</p>
+					<p ng-if="feature.learnMoreURL">Learn more at <a ng-href="{{ feature.learnMoreURL }}">{{ feature.learnMoreURL }}</a>.</p>
+	        <div class="image-div">
+						<img ng-if="feature.img" ng-src="{{feature.img}}">
+					</div>
+				</div>
+      </div>
+    </div>
+		<br><br><br><br><br>
```
timlevett: Holy `<br>` batman

vertein: styling in the html code is where it's at

src/main/webapp/portal/main/controllers.js
```diff
@@ -61,27 +61,29 @@ define(['angular','require'], function(angular, require) {
...
+            if (today > startDate && today < endDate) {
               $modal.open({
                 animation: $scope.animationsEnabled,
-                templateUrl: require.toUrl('./partials/welcome-modal-template.html'),
+                templateUrl: require.toUrl('./partials/features-modal-template.html'),
                 size: 'lg',
                 scope: $scope 
               });
-              $localStorage.hasSeenWelcome = true;
+              $localStorage.hasSeenLatestFeature = true;
```
timlevett: Discussed f2f, end result, lets store the last id

apetro:
Agreeing with the suggestions above, and :+1:.

src/main/webapp/portal/main/controllers.js
jhanstra: Changed and tested, users should see new features added now.

src/main/webapp/portal/features/partials/features.html
jhanstra: Yea I suppose that's bad practice. I'm not a huge fan of one-off margin styles either, but it's probably better than <br>s :P

PR #15, opened 2015-08-25
Provides more detail in README

nblair:
This pull request adds some more detail to the README for prospective consumers. Displays the maven dependency and references the main.js extension point.

Not sure if I got the image markdown correct, we'll see how it looks.


apetro:
:+1: .  [Looks like](https://github.com/nblair/uw-frame/blob/readme-help/README.md) the image markdown is just fine.

vertein:
What it looks like:

image

image

PR #16, opened 2015-08-25
MUMUP-2060:: Clean splash seen-ness

jhanstra:
Logic for determining whether to show the popup to a user and handling legacy local storage variables. @timlevett does this cover every use case you were thinking of?

### DIFF
--- a/src/main/webapp/portal/main/controllers.js
+++ b/src/main/webapp/portal/main/controllers.js
@@ -84,6 +84,7 @@ define(['angular','require'], function(angular, require) {
               } else {
                 $localStorage.lastSeenFeature = 0;
               }
+              delete $localStorage.hasSeenWelcome;
             }
             
             // criteria to show popup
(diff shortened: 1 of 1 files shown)


src/main/webapp/portal/main/controllers.js
```diff
@@ -76,7 +76,23 @@ define(['angular','require'], function(angular, require) {
...
             var endDate = Date.parse(new Date($scope.latestFeature.popup.endYear, $scope.latestFeature.popup.endMonth, $scope.latestFeature.popup.endDay));
-            if ((today > startDate && today < endDate) && $localStorage.lastSeenFeature < $scope.latestFeature.id) {
+            
+            // handle legacy local storage
+            if ($localStorage.lastSeenFeature === -1) {
+              if ($localStorage.hasSeenWelcome) {
+                $localStorage.lastSeenFeature = 1;
+              } else {
+                $localStorage.lastSeenFeature = 0;
+              }
```
timlevett: Only suggestion would be do a `delete $localStorage.hasSeenWelcome;` after the if/else block. #cleanup

jhanstra: Good idea

timlevett:
:+1:

PR #17, opened 2015-08-27
MUMUP-2078 Home : aria-landmark role usages -[uw-frame]

jiayinjx:
Add aria-landmark role usages in support of accessibility.


timlevett:
Overall neat! :+1: . Thoughts on adding a `role="main"` to where the ng-view is in the frame.jsp?

jiayinjx:
will do.

vertein:
:+1:

PR #21, opened 2015-09-03
MUMUP-2087 : cleanup features null state

timlevett:
+ Set default to `[]`
+ Only do stuff if stuff to do
+ Disable by default


vertein:
:+1:

PR #24, opened 2015-09-15
MUMUP-2110 : Fix empty string issue with value, add in test

timlevett:
+ Changed behavior so instead of empty string it'll return an empty array on an empty state of the notification:dismiss KV.
+ Created test case to make sure that stays that way
+ Documentation for storage service


vertein:
:+1:

PR #29, opened 2015-09-24
MUMUP-210[7-8] : Hide flags for hiding things

timlevett:
+ `APP_FLAG.hideSidebar` : Hides sidebar and sidebar pill. Also sets default sidebar to hidden so it'll hide
+ `APP_FLAG.hideSearch` : Hides search bar in header

I think you can imagine what that looks like.


jhanstra:
Cool :+1:

PR #32, opened 2015-09-30
Cleanup v2

### LINKED ISSUE #30: Namespace Issues
I'm porting a stand alone angular-based application to live in the uw-frame and had previously made the unfortunate choice to name one of the controllers 'SearchController'.  

I eventually realized that the uw-frame angular app already registered a controller constructor named 'SearchController', and in this case, the uw-frame's version was clobbering mine, since it happened to register after my version. I solved my problem by simply registering it under a different name.

To avoid confusion, I would suggest registering all of the built-in uw-frame components with a common prefix, such as 'Portal', including controllers, service, directives and filters, since any of them could be candidates for namespace collisions. Angular constructor functions (and their names) get put in the same application-global bucket for that type of component, even across angular modules.

To be clear, this isn't about module names or RequireJS issues, but the name that is used when a component constructor is registered with angular, like this:

```javascript
angular.module('portal.search.controllers').controller('SearchController',['blah',function(blah) { 
}]);
```
This would be better:
```javascript
angular.module('portal.search.controllers').controller('PortalSearchController',['blah',function(blah) { 
}]);
```
Without a prefixing scheme of some kind, this issue will probably keep coming up as both the uw-frame and frame apps evolve over time.

If I missed some other way to avoid this problem, feel free to to enlighten me about it. :)

timlevett:
+ Features cleanup. Added empty state, changed `MyUW` to `{{NAMES.title}}`, and polished controller.
+ Namespaced a bunch of controllers and a few services (prefix with Portal). This is due to angular having global scope on controller names. Should resolve #30 
+ negated the hide[Sidebar|Search] to be show[Sidebar|Search] to avoid double negative. This requires a minor release so bumped the version.

### DIFF
--- a/src/main/webapp/portal/main/controllers.js
+++ b/src/main/webapp/portal/main/controllers.js
@@ -62,12 +62,12 @@ define(['angular','require'], function(angular, require) {
     });
   }]);
   
-  app.controller('PopupController', ['$localStorage', '$sessionStorage','$scope', '$document', 'APP_FLAGS', '$modal', 'featuresService', '$sanitize', function($localStorage, $sessionStorage, $scope, $document, APP_FLAGS, $modal, featuresService, $sanitize) {
+  app.controller('PopupController', ['$localStorage', '$sessionStorage','$scope', '$document', 'APP_FLAGS', '$modal', 'portalFeaturesService', '$sanitize', function($localStorage, $sessionStorage, $scope, $document, APP_FLAGS, $modal, portalFeaturesService, $sanitize) {
      var openModal = function() {
       if (APP_FLAGS.features) {
         $scope.features = [];
         
-        featuresService.getFeatures().then(function(data) {
+        portalFeaturesService.getFeatures().then(function(data) {
             var features = data;
             if (features.data.length > 0) {
               $scope.features = features.data;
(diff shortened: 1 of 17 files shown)


vertein:
:+1:

jhanstra:
:+1: for namespacing

src/main/webapp/portal/main/controllers.js
```diff
@@ -3,7 +3,7 @@
 define(['angular','require'], function(angular, require) {
   var app = angular.module('portal.main.controllers', []);
 
-  app.controller('MainController', ['$localStorage', '$sessionStorage','$scope', '$document', 'NAMES', 'MISC_URLS', 'APP_FLAGS', function($localStorage, $sessionStorage, $scope, $document, NAMES, MISC_URLS, APP_FLAGS) {
+  app.controller('PortalMainController', ['$localStorage', '$sessionStorage','$scope', '$document', 'NAMES', 'MISC_URLS', 'APP_FLAGS', function($localStorage, $sessionStorage, $scope, $document, NAMES, MISC_URLS, APP_FLAGS) {
     var defaults = {
             showSidebar: !APP_FLAGS.hideSidebar,
```
jhanstra: Should this say `showSidebar: APP_FLAGS.showSidebar' now?

timlevett: good catch!

timlevett: Fixed https://github.com/UW-Madison-DoIT/uw-frame/commit/926d0b386c91ee8efbd922f33c907ba2473ab98a

PR #33, opened 2015-09-30
MUMUP-1926: Add google analytics tracking for sidebar events

jhanstra:
Hopefully this will give us better insights into user behavior regarding the sidebar - whether people use it, how often they minimize it and maximize it, which links they are clicking in the navigation...

@timlevett, did I use your nifty `pushGaEvent()` function correctly? Is there a way I can test this locally?

### DIFF
--- a/src/main/webapp/frame.jsp
+++ b/src/main/webapp/frame.jsp
@@ -51,7 +51,7 @@
           <div class="region-sidebar-left col-sm-2 col-xs-0 hidden-xs no-margin" ng-if="$storage.showSidebar && APP_FLAGS.showSidebar">
               <side-bar-menu></side-bar-menu>
           </div>
-          <div ng-if="(!($storage.showSidebar)) && APP_FLAGS.showSidebar" class="show-sidebar" ng-click="$storage.showSidebar = true">
+          <div ng-if="(!($storage.showSidebar)) && APP_FLAGS.showSidebar" class="show-sidebar" ng-click="$storage.showSidebar = true" onClick="pushGaEvent('Sidebar','Maximize Sidebar', 'Maximize Sidebar')">
             <span class="fa fa-bars"></span>
           </div>
           <div role="main" id="region-main" class="col-xs-12" ng-class="{'col-sm-10 col-sm-offset-2' : ($storage.showSidebar && APP_FLAGS.showSidebar), 'col-sm-11 max-view' : (!($storage.showSidebar) || !APP_FLAGS.showSidebar)}">
--- a/src/main/webapp/frame.jsp
+++ b/src/main/webapp/frame.jsp
@@ -51,7 +51,7 @@
           <div class="region-sidebar-left col-sm-2 col-xs-0 hidden-xs no-margin" ng-if="$storage.showSidebar && APP_FLAGS.showSidebar">
               <side-bar-menu></side-bar-menu>
           </div>
-          <div ng-if="(!($storage.showSidebar)) && APP_FLAGS.showSidebar" class="show-sidebar" ng-click="$storage.showSidebar = true" onClick="pushGaEvent('Sidebar','Maximize Sidebar', 'Maximize Sidebar')">
+          <div ng-if="(!($storage.showSidebar)) && APP_FLAGS.showSidebar" class="show-sidebar" ng-click="openSidebar()">
             <span class="fa fa-bars"></span>
           </div>
           <div role="main" id="region-main" class="col-xs-12" ng-class="{'col-sm-10 col-sm-offset-2' : ($storage.showSidebar && APP_FLAGS.showSidebar), 'col-sm-11 max-view' : (!($storage.showSidebar) || !APP_FLAGS.showSidebar)}">
--- a/src/main/webapp/js/app-config.js
+++ b/src/main/webapp/js/app-config.js
@@ -27,7 +27,7 @@ define(['angular'], function(angular) {
             'searchURL' : 'https://my.wisc.edu/web/apps/search/'
         })
         .constant('NOTIFICATION', {
-            'enabled' : false,
+            'enabled' : true,
             'groupFiltering' : false,
             'notificationFullURL' : 'notifications'
         })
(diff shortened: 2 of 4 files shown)


keirserrie:
:+1:

src/main/webapp/portal/main/partials/sidebar-left.html
```diff
@@ -1,7 +1,7 @@
 <div>
   <nav role="navigation" class="side-bar-nav">
     <div class="hide-sidebar-div">
-      <span class="fa fa-arrow-left hide-sidebar" ng-click="$storage.showSidebar = false" popover="Minimize sidebar" popover-placement="bottom" popover-trigger="mouseenter" popover-popup-delay="600" popover-append-to-body="true"></span>
+      <span class="fa fa-arrow-left hide-sidebar" ng-click="$storage.showSidebar = false" popover="Minimize sidebar" popover-placement="bottom" popover-trigger="mouseenter" popover-popup-delay="600" popover-append-to-body="true" onClick="pushGaEvent('Sidebar','Minimize Sidebar', 'Minimize Sidebar')"></span>
```
apetro: Subtle tweak: would it be better to for the maximize and minimize events to share an action and differentiate by label, so as to more easily group and compare the maximizing and minimizing?

`"pushGaEvent('Sidebar','Show-hide Sidebar', 'Show Sidebar')"`
`"pushGaEvent('Sidebar','Show-hide Sidebar', 'Hide Sidebar')"`

apetro:
:+1:

apetro:
This will fire GA events when users show and hide sidebar. But it won't tell us what state the sidebar is in when people render the page and don't touch it.  Should there be a fire-on-render event along the lines of

`"pushGaEvent('Sidebar','Rendered', 'Expanded')"`
`"pushGaEvent('Sidebar','Rendered', 'Minimized')"`

to give visibility into not just how much people change the state how, but what state it's in when people are using the portal?  Or that'll be too much events overload?

src/main/webapp/frame.jsp
```diff
@@ -51,7 +51,7 @@
           <div class="region-sidebar-left col-sm-2 col-xs-0 hidden-xs no-margin" ng-if="$storage.showSidebar && APP_FLAGS.showSidebar">
               <side-bar-menu></side-bar-menu>
           </div>
-          <div ng-if="(!($storage.showSidebar)) && APP_FLAGS.showSidebar" class="show-sidebar" ng-click="$storage.showSidebar = true">
+          <div ng-if="(!($storage.showSidebar)) && APP_FLAGS.showSidebar" class="show-sidebar" ng-click="$storage.showSidebar = true" onClick="pushGaEvent('Sidebar','Maximize Sidebar', 'Maximize Sidebar')">
```
timlevett: This won't work as `onClick` wouldn't have the context of Angular. You will need to do a `ng-click` and add a function to the controller.  Possibly a "toggleSidebar" function, which just fires the proper GA event and switches the `$storage.showSidebar`.

src/main/webapp/portal/main/partials/sidebar-left.html
timlevett: Agreed with suggestions. See note above about switching function

nblair:
We are using this angular module in DoIT Number:

https://luisfarzati.github.io/angulartics/

It has handy element attributes for doing what you want.

src/main/webapp/frame.jsp
jhanstra: @timlevett Is there any distinct advantage to using the angular pushGaEvent method in the `miscService` over just using something like `onClick="_gaq.push(['_trackEvent', 'Beta', 'Read About Change', 'Read About Change']);"`? I can't remember if we talked about this. Seems like this would be much easier than pulling in the dependency anywhere we need to track events.

timlevett:
@nblair - that may be nice if we decide to revamp the analytics module, which is very much out of scope of this little issue. ;) i'll backlog an issue as people seems bothered by our current approach.

jhanstra:
Agreed with @timlevett. We can revisit angulartics if/when we're ready to revamp analytics.

jhanstra:
It should work now.
image

src/main/webapp/js/app-config.js
```diff
@@ -27,7 +27,7 @@ define(['angular'], function(angular) {
             'searchURL' : 'https://my.wisc.edu/web/apps/search/'
         })
         .constant('NOTIFICATION', {
-            'enabled' : false,
+            'enabled' : true,
```
timlevett: This should be `false` by default.

src/main/webapp/portal/main/partials/sidebar-left.html
```diff
@@ -11,8 +11,8 @@
     <hr>
     <div ng-controller="PortalSidebarController as sidebarCtrl">
         <ul class="list-group sidebar-list">
-          <li class="list-group-item" ng-repeat="item in sidebar" ng-if="sidebarCtrl.canSee(item.displayFlag)">
-            <a href="{{item.hyperlink}}" target="{{item.target}}" ng-click="headerCtrl.navbarCollapsed = true">
+          <li class="list-group-item" ng-repeat="item in sidebar" ng-if="sidebarCtrl.canSee(item.displayFlag)" ng-click="headerCtrl.navbarCollapsed = true;">
+            <a href="{{item.hyperlink}}" target="{{item.target}}" ng-click="pushGAEvent('Sidebar', 'Click Link', item.title);">
```
timlevett: Neat!

timlevett:
1 fix in `app-config.js`, then I will thumb.

src/main/webapp/js/app-config.js
jhanstra: hmm not sure how that changed

vertein: Don't play coy with us, @jhanstra

jhanstra: oh wait, that was from when I was testing notifications way back when. never mind, move along.

timlevett:
:+1:

PR #36, opened 2015-10-09
Add helpdeskURL to MISC_URLS.

apetro:
Add a `helpdeskURL` to the `MISC_URLS` map.

(Intended use: showing link among no-search-results recovery options when this URL is configured.)


timlevett:
:-1: This URL is not going to be used anywhere in frame, therefore just put it in `angularjs-portal`'s `app-config.js`

mrfarnham:
Can't remember if the coarse grained authorization error page shows the
help desk link or not.

Was also thinking add a url for the outages page so it could be displayed
for an error message or status page.

PR #38, opened 2015-10-09
Fix notificationURL being null throwing errors

### LINKED ISSUE #35: If notificationURL is null, don't run the service
Please add a check to notification service so if `SERVICE_LOC.notificationURL` is null or undefined (falsy) it doesn't run the service. Its causing an error.

timlevett:
When `SERVICE_LOC.notificationURL` was `null` (regardless if `NOTIFICATION.enabled` was `true` or `false`) it would throw an protocol error.

This "fixes" that so it only throws that if you have `NOTIFICATION.enabled` set to `true`.

This resolves #35


.codenvy/runners/environments/Copy of Tomcat 7.0  Java 7/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+#
+
+FROM codenvy/jdk7_tomcat7
+
+$debug?EXPOSE 8000:$
+$debug?ENV CODENVY_APP_PORT_8000_DEBUG 8000:$
+$debug?CMD ./catalina.sh jpda run 2>&1:$
+
+ENV CODENVY_APP_BIND_DIR /home/user/tomcat7/webapps/ROOT
+VOLUME ["/home/user/tomcat7/webapps/ROOT"]
```
timlevett: Not a big deal having this, but could remove if others dislike

apetro:
:+1: .

PR #39, opened 2015-10-09
MUMUP-2148

jiayinjx:
update uw-frame, rename "samples" to "dynamicFeeds" and all corresponding files in uw-frame


timlevett:
Rename to `staticFeeds` like in `angularjs-portal`

PR #42, opened 2015-10-13
MUMUP-2013 : notifications in xs mode

timlevett:
+ Add `mode` to the `<notification-bell>` directive
+ `mode='mobile-menu'` will result in    
                  
http://goo.gl/BrzrCY
+ `mode='bell'` and it'll show the bell `(used in the desktop nav bar)`
http://goo.gl/pvWCqH
+ `mode='mobile-bell'` will result in a cute little notification badge on your burger menu `iff unread count > 0)`

http://goo.gl/k2YnoM
+ Also fixed GA pageview with using mode. If `directiveMode` is not set in the notification controller (which it won't be on `notification-full.html`) then it'll do a page hit, otherwise it just carry's on.
+ I also (cough) shrunk the header in `notifications-full.html`

http://goo.gl/17AwVG

Based on mockups from Keir that are in the MUM > Notification folder of Google Drive

### DIFF
--- a/src/main/webapp/css/buckyless/notifications.less
+++ b/src/main/webapp/css/buckyless/notifications.less
@@ -125,10 +125,6 @@
     color: @color1;
 }
 
-i.fa.fa-bell.fa-2x.notification-badge-mobile {
-    font-size: smaller;
-}
-
 //mobile menu bell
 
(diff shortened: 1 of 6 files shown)


vertein:
:+1:

PR #43, opened 2015-10-14
MUMUP-2013 : Notification xs fix multiple service calls

timlevett:
+ Fix issue with notification dismiss service getting called 3 times during init page load

#### Before
image

#### After
image

+ Fix issue with clicking on the menu item for notification (in xs) and the menu not hiding after partial loading.

#### Before
http://goo.gl/nErXgr

#### After
http://goo.gl/RNcw0v


vertein:
:+1:

PR #45, opened 2015-10-15
MUMUP-2127 : About partial

timlevett:
+ Adds `/about` page
+ Adds a `about-frame.json` file that is filtered via maven during build to get build information
+ Adds `frame-config.js` file to put in the URL specific for frame information. Apps shouldn't overwrite that so therefore frame finally gets its own config file.
+ Adds `aboutURL` to the `SERVICE_LOC` constant. This is to get information about the application. This is optional.

#### Screenshots

With app info
http://goo.gl/M7hOxV
_Note info in app info section is just the frame info for example_

Without app info
http://goo.gl/NeimIk


jhanstra:
Nice :+1:

PR #47, opened 2015-10-16
Remove duplicate styles that are now included in uw-ui-toolkit

timlevett:


timlevett:
Self approving as it fixes some issues :+1:

PR #48, opened 2015-10-20
Minor cleanup of guest mode for more generic guest username

timlevett:
+ Moves guest username to constants
+ Check if user object exists before accessing it.


nblair:
:+1:

PR #49, opened 2015-10-21
Hide '-' spans for serverName and/or sessionKey if not set

nblair:
Some apps override `SERVICE_LOC.sessionInfo` the URL to customize the name in the side bar and build version in the footer. 
We don't know the serverName, nor do we have the portal sessionKey, so the JSON response for our custom URL doesn't include those values.

In the current uw-frame release, this results in the dashes between those fields rendering, and looking pretty funny:

screen shot 2015-10-21 at 12 59 19 pm

This pull request makes a small tweak to the footer to only show those dashes when the respective fields are set. Results in no more funny looking dashes:

screen shot 2015-10-21 at 1 11 57 pm

### DIFF
--- a/src/main/webapp/portal/main/partials/footer.html
+++ b/src/main/webapp/portal/main/partials/footer.html
@@ -12,7 +12,7 @@
 				<span><a ng-if='sessionCheckCtrl.back2ClassicURL' ng-href='{{sessionCheckCtrl.back2ClassicURL}}'>MyUW Classic</a></span>
 			</div>
 			<div>
-				<span>{{sessionCheckCtrl.user.version}}</span><span ng-show="sessionCheckCtrl.user.serverName"> - </span><span>{{sessionCheckCtrl.user.serverName}}</span><span ng-show="sessionCheckCtrl.user.sessionKey"> - </span><span>{{sessionCheckCtrl.user.sessionKey}}</span></span>
+				<span>{{sessionCheckCtrl.user.version}}</span><span ng-show="sessionCheckCtrl.user.serverName"> - {{sessionCheckCtrl.user.serverName}}</span><span ng-show="sessionCheckCtrl.user.sessionKey"> - {{sessionCheckCtrl.user.sessionKey}}</span></span>
 			</div>
 		</div>
 	</div>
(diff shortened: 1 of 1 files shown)


src/main/webapp/portal/main/partials/footer.html
```diff
@@ -12,7 +12,7 @@
 				<span><a ng-if='sessionCheckCtrl.back2ClassicURL' ng-href='{{sessionCheckCtrl.back2ClassicURL}}'>MyUW Classic</a></span>
 			</div>
 			<div>
-				<span>{{sessionCheckCtrl.user.version}}</span><span> - </span><span>{{sessionCheckCtrl.user.serverName}}</span><span> - </span><span>{{sessionCheckCtrl.user.sessionKey}}</span></span>
+				<span>{{sessionCheckCtrl.user.version}}</span><span ng-show="sessionCheckCtrl.user.serverName"> - </span><span>{{sessionCheckCtrl.user.serverName}}</span><span ng-show="sessionCheckCtrl.user.sessionKey"> - </span><span>{{sessionCheckCtrl.user.sessionKey}}</span></span>
```
paulerickson: That's a lot of spans for one line!  Could this be trimmed down to:

```
<span>{{sessionCheckCtrl.user.version}}</span>
<span ng-show="sessionCheckCtrl.user.serverName"> - {{sessionCheckCtrl.user.serverName}}</span>
<span ng-show="sessionCheckCtrl.user.sessionKey"> - {{sessionCheckCtrl.user.sessionKey}}</span>
```

nblair: Agreed! Wasn't sure why it was there, I'll see if it works with this alternative.

nblair: Might want My UW folks to weigh in; @jhanstra or @vertein do you have anything looking for these specific values in dedicated spans? They don't have ids or classes, so I'm betting not.

A local test of Paul's technique shows it works.

vertein: You're right.  I can't find any specific stylings.

vertein:
I like @paulerickson idea of less spans

vertein:
And now, I see you did the commit

vertein:
:+1:

PR #50, opened 2015-10-29
MUMUP-2207 Tweaks the /web-portlet-rendering flavor text.

apetro:
Reminds that the feature works only for compact-mode widget cards.


timlevett:
:+1:

timlevett:
One may suggest we need a `/web` specific settings page :)

PR #52, opened 2015-10-29
Make Home and Log out non-overridable via sidebar

### LINKED ISSUE #51: 'Log out' URL defined in 2 places
In 1.2.0 release, there are 2 configuration locations for a 'Log out' URL:

1. in staticFeeds/sidebar.json.
2. in js/app-config.js (`MISC_URLS.logoutURL`)

Is there any way those can be consolidated?

nblair:
Resolves #51.

Prior to this change set, the links for Home and Log out in the sidebar came from the array returned by the URL defined in `SERVICE_LOC.sidebarInfo`. 

In this change set, the Home and Log out links in the sidebar are always there; Home is hard set to '/' (with target="_self", so it's not the route /, but the real path /), and Log out is configured via `MISC_URLS.logoutURL`.

A new horizontal rule appears below, and links are created from the array returned by the URL defined in `SERVICE_LOC.sidebarInfo` underneath.

In action:

screen shot 2015-10-29 at 2 21 36 pm

### DIFF
--- a/src/main/webapp/portal/main/partials/sidebar-left.html
+++ b/src/main/webapp/portal/main/partials/sidebar-left.html
@@ -22,6 +22,7 @@
             </a>
           </li>
        </ul>
+       <div ng-show="sidebar.length">
        <hr/>
         <ul class="list-group sidebar-list">
           <li class="list-group-item" ng-repeat="item in sidebar" ng-if="sidebarCtrl.canSee(item.displayFlag)" ng-click="headerCtrl.navbarCollapsed = true;">
@@ -30,6 +31,7 @@
             </a>
           </li>
        </ul>
+       </div
    </div>
    <ul class="list-group sidebar-list" >
     <div class="quicklinks" ng-if="$storage.sidebarQuicklinks">
--- a/src/main/webapp/portal/main/partials/sidebar-left.html
+++ b/src/main/webapp/portal/main/partials/sidebar-left.html
@@ -10,6 +10,19 @@
     </div>
     <hr>
     <div ng-controller="PortalSidebarController as sidebarCtrl">
+    	<ul class="list-group sidebar-list">
+          <li class="list-group-item" ng-click="headerCtrl.navbarCollapsed = true;">
+            <a href="/" target="_self" ng-click="pushGAEvent('Sidebar', 'Click Link', 'Home');">
+              <span class="fa fa-home fa-fw"></span> Home
+            </a>
+          </li>
+          <li class="list-group-item" ng-click="headerCtrl.navbarCollapsed = true;">
+            <a href="{{MISC_URLS.logoutURL}}" target="_self" ng-click="pushGAEvent('Sidebar', 'Click Link', 'Log out');">
+              <span class="fa fa-sign-out fa-fw"></span> Log out
+            </a>
+          </li>
+       </ul>
+       <hr/>
         <ul class="list-group sidebar-list">
           <li class="list-group-item" ng-repeat="item in sidebar" ng-if="sidebarCtrl.canSee(item.displayFlag)" ng-click="headerCtrl.navbarCollapsed = true;">
             <a href="{{item.hyperlink}}" target="{{item.target}}" ng-click="pushGAEvent('Sidebar', 'Click Link', item.title);">
--- a/src/main/webapp/staticFeeds/sidebar.json
+++ b/src/main/webapp/staticFeeds/sidebar.json
@@ -1,7 +1,7 @@
 {"sidebar" :  
             [
-                {"title": "Home", "hyperlink" : "/", "target" : "", "faIcon" : "fa-home", "displayFlag" : ""},
                 {"title": "Beta Settings", "hyperlink" : "settings", "target" : "", "faIcon" : "fa-cog", "displayFlag" : "sidebarShowSettings"},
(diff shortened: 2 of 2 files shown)


timlevett:
:+1: Neat! Nice work @nblair . Thanks for the contribution.

jhanstra:
:+1:

nblair:
Should we get @keirserrie to comment?

timlevett:
Yeah, I was going to say that, took the words right out of my mind. @keirserrie ?

src/main/webapp/staticFeeds/sidebar.json
```diff
@@ -1,7 +1,7 @@
 {"sidebar" :  
             [
-                {"title": "Home", "hyperlink" : "/", "target" : "", "faIcon" : "fa-home", "displayFlag" : ""},
                 {"title": "Beta Settings", "hyperlink" : "settings", "target" : "", "faIcon" : "fa-cog", "displayFlag" : "sidebarShowSettings"},
-                {"title": "Log out", "hyperlink" : "/portal/Logout", "target" : "", "faIcon" : "fa-sign-out", "displayFlag" : ""}
+                {"title": "External URL", "hyperlink" : "http://www.wisc.edu", "target" : "", "faIcon" : "fa-external-link", "displayFlag" : ""},
```
apetro: External URL should have a `target` of  `_blank`,  just so the example is supporting the pattern.

src/main/webapp/portal/main/partials/sidebar-left.html
```diff
@@ -10,6 +10,19 @@
...
+              <span class="fa fa-home fa-fw"></span> Home
+            </a>
+          </li>
+          <li class="list-group-item" ng-click="headerCtrl.navbarCollapsed = true;">
+            <a href="{{MISC_URLS.logoutURL}}" target="_self" ng-click="pushGAEvent('Sidebar', 'Click Link', 'Log out');">
+              <span class="fa fa-sign-out fa-fw"></span> Log out
+            </a>
+          </li>
+       </ul>
+       <hr/>
```
apetro: Condition the `<hr/>` on there being at least one additional sidebar item that the viewer `canSee` ?

apetro:
This changes the semantics of `sidebar.json`, requiring `frame`-based apps to update their `sidebar.json` when upgrading to a `uw-frame` release with this change.  Therefore bump [the version number](https://github.com/UW-Madison-DoIT/uw-frame/blob/2ca1ef422556d8f1f1b81030e94089acff2c4fc5/pom.xml#L7) [semantically](http://semver.org/)?

src/main/webapp/portal/main/partials/sidebar-left.html
```diff
@@ -10,6 +10,19 @@
...
         <ul class="list-group sidebar-list">
           <li class="list-group-item" ng-repeat="item in sidebar" ng-if="sidebarCtrl.canSee(item.displayFlag)" ng-click="headerCtrl.navbarCollapsed = true;">
```
apetro: If this `ng-repeat` (or `ng-if`) were to filter out `sidebar` items titled `"Home"` and `"Log out"`, then the `frame` with this changeset could be upgraded to with less update by inheriting `frame`-based applications, that is, they would have less need to update their `sidebar.json`.  Might not be worth the complexity cost to achieve a less-change migration path.

src/main/webapp/staticFeeds/sidebar.json
nblair: It technically doesn't have to; links that have scheme/host/port aren't handled by the angular router.

src/main/webapp/portal/main/partials/sidebar-left.html
nblair: :+1:

src/main/webapp/portal/main/partials/sidebar-left.html
nblair: I'd rather not have that complexity; from what I gather there is strong desire for Home and Log out to always be present and not change behavior, might as well formalize that by making them first class elements in the markup rather than generated via customizable code.

src/main/webapp/portal/main/partials/sidebar-left.html
nblair: Fixed with 5fdd6aa2.

nblair:
Screenshot showing hr disappearing for empty sidebarInfo:

screen shot 2015-10-29 at 8 38 31 pm

src/main/webapp/staticFeeds/sidebar.json
apetro: Ah, I wasn't clear.  It should have a `target` of `_blank`, because it is functioning as an example for downstream (frame-based-app) developers, and the UX pattern in MyUW is that `href`s to external URLs use `target="_blank"`.  *should* in the `RFC` sense of the word, as it were.

nblair: Ah right, I forgot about that :+1:

nblair:
`target="_blank"` now set for the external url example. It's a shame JSON has no support for inline comments, there's definitely some tricks to know (like this convention, and _self for non-angular routes, and the how the displayFlag works).

nblair:
Regarding versioning, would you want to see the version increment happen on this branch before merging?

timlevett:
We usually just flag the change with the label that requests a minor version change. Marking.

keirserrie:
Sorry so late here, this is fine with the caveat that the sidebar may be going away...waiting on analytics before the final call is made.

PR #56, opened 2015-11-02
Log out link for showSidebar=false needs different 'target' attribute

nblair:
Needs `target='_self'` because the link is pretty much guaranteed to not be an angular route.


timlevett:
:+1:

PR #58, opened 2015-11-03
[WIP] Dismissed notification bucket

timlevett:
Thought brought up in MUMMNG-2071. What do we do with notifications that are no longer relevant?  Here is one take. We move them to another bucket in the notification json.

image

### DIFF
--- a/src/main/webapp/portal/notifications/partials/notifications-full.html
+++ b/src/main/webapp/portal/notifications/partials/notifications-full.html
@@ -10,19 +10,29 @@
 		</div>
 		<div>
 			<ul class="notifications-list">
-        		<li class="notification-item" ng-repeat="notification in notifications" ng-if='shouldShow(notification)'>
+        <li class="notification-item" ng-repeat="notification in notifications" ng-if='shouldShow(notification)'>
 					<a class="content" ng-href='{{notification.actionURL}}' target='_blank' aria-label='{{notification.actionAlt}}'>
-	            		<p>{{ notification.title }}</p>
+	          <p>{{ notification.title }}</p>
 					</a>
 					<div class="action" ng-click='dismiss(notification)' ng-hide="!notification.dismissable || isDismissed(notification)">
-	            	<i class='fa fa-times'tooltip="Dismiss notification" tooltip-placement="top" tooltip-popup-delay="300" tooltip-trigger="mouseenter"></i>
+	          <i class='fa fa-times'tooltip="Dismiss notification" tooltip-placement="left" tooltip-popup-delay="300" tooltip-trigger="mouseenter"></i>
 					</div>
-               <div class="action" ng-click="undismiss(notification)" ng-hide="!(mode === 'old')">
-                 <i class='fa fa-undo'tooltip="Revert Dismiss of notification" tooltip-placement="left" tooltip-popup-delay="300" tooltip-trigger="mouseenter"></i>
-              </div>
-        		</li>
-        		<li class="notification-item read" ng-if="isEmpty || (countWithDismissed() === 0 && mode === 'new') || (mode === 'old' && dismissedCount() === 0)">You have no <span ng-show='dismissedCount() === 0'>dismissed</span> notifications at this time.</li>
-        	</ul>
+          <div class="action" ng-click="undismiss(notification)" ng-hide="!(mode === 'old')">
+             <i class='fa fa-undo'tooltip="Revert Dismiss of notification" tooltip-placement="left" tooltip-popup-delay="300" tooltip-trigger="mouseenter"></i>
+          </div>
+        </li>
+        <li class="notification-item read" ng-if="isEmpty || (countWithDismissed() === 0 && mode === 'new') || (mode === 'old' && dismissedCount() === 0 && dismissedNotifications.length === 0)">You have no <span ng-show='dismissedCount() === 0'>dismissed</span> notifications at this time.</li>
+      </ul>
+			<ul class="notifications-list" ng-if="mode === 'old'">
+				<li class="notification-item" ng-repeat="notification in dismissedNotifications">
+					<p class='content'>{{ notification.title }}</p>
+					<div class='action'>
+						<i class='fa fa-info' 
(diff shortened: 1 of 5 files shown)


keirserrie:
:+1:  So this gives us expiration on notifications? Also, hover on "i" icon, would be tap on mobile?

timlevett:
@keirserrie : this gives us the illusion of expired notifications. Not sure on the mobile thing. I'll have to check that out when it gets to an env. Could switch to a click if need be.

src/main/webapp/staticFeeds/sample_notifications.json
```diff
@@ -24,5 +24,19 @@
         "actionAlt" : "Activate Services",
         "dismissable" : true
       }
-    ]
+    ],
+  "dismissedNotifications" : [
```
apetro: Pros/cons on a separate `dismissedNotifications` array vs a single array with items having a `dismissedDate` of today or earlier being automatically dismissed?  The latter approach moves parsing cost to run-time (rather than deploy-time pre-processing of this data into the different arrays), but it picks up the feature of being able to schedule out auto-dismiss into the future at the time the notification is published, no different `JSON` to publish to switch the notification state on the intended auto-dimiss-day.

timlevett: Valid ideas @apetro . Had similar thoughts. I like the idea. Note that this does complicate calculating the number of new notifications. That would be my only concern, as we would have to do date compares to see if its new.

src/main/webapp/portal/notifications/partials/notifications-full.html
```diff
@@ -10,19 +10,29 @@
...
+          </div>
+        </li>
+        <li class="notification-item read" ng-if="isEmpty || (countWithDismissed() === 0 && mode === 'new') || (mode === 'old' && dismissedCount() === 0 && dismissedNotifications.length === 0)">You have no <span ng-show='dismissedCount() === 0'>dismissed</span> notifications at this time.</li>
+      </ul>
+			<ul class="notifications-list" ng-if="mode === 'old'">
+				<li class="notification-item" ng-repeat="notification in dismissedNotifications">
+					<p class='content'>{{ notification.title }}</p>
+					<div class='action'>
+						<i class='fa fa-info' 
+						   tooltip='dismissed {{notification.dismissedDate.m + "-" + notification.dismissedDate.d + "-" + notification.dismissedDate.y}}'
```
apetro: "Dismissed" sounds like something the user does. This tooltip messaging may be confusing in that it's telling the user that a notification was dismissed on a date, when the user didn't necessarily dismiss the notification on that date or take any action on the notification at all, indeed may have never seen this notification before.  Also, if these tooltips show when these notifications were dismissed, the user might reasonably wonder, okay, but what about those other dismissed notifications up there with the revert icon thing, when did I dismiss those?  

So maybe this "administratively dismissed" state would be better called something else.  "Expired", perhaps?  "Expired 10-31-2015" helps imply why the user can't un-dismiss that one, differentiating if from merely dismissed items that can be un-dismissed?

apetro:
:+1: .  I love this feature.  As noted inline, there are some details of the implementation and of the user interactions that are, well, complicated.

We have four-ish notifications today, for small values of four.  I'm fine with shipping this feature as-is and learning from it.  I'm also fine with giving some of these details some more thought and tweaking this some before shipping it.

jhanstra:
Agreed with @apetro that there's a subtle difference between a user dismissing a notification and us automatically 'dismissing' it for irrelevance. We will probably want to think about addressing that nuance in the future. But I'm also a :+1: for :shipit:

src/main/webapp/portal/notifications/partials/notifications-full.html
timlevett: The problem with creating yet another bucket is it makes things complex (too complex, debatable). Then you have these odd edge cases like "person dismissed, but then notification expires, where does it go?". The answer is, it could go in both. I like the 2 states we have now, I feel we would complicate things by adding another.

timlevett:
Not shipping as is, but got some great ideas. I think we will revisit this with a better step forward. You will see rev 2 during our next lul

src/main/webapp/portal/notifications/partials/notifications-full.html
apetro: Yeah. I suspect it's not fully another bucket, it's a nuance on how we talk about and treat the notifications in the dismissed bucket. Some can be un-dismissed.  Some cannot. What handles can we give the user to understand why? Well. It turns out it depends on how the notification got into that dismissed bucket. If the user put it there, it's dismissed, and can be un-dismissed.  But if we put it there, it's expired, and the user can't un-expire it.  Expired isn't a different place, a different bucket -- it's a different state of mind in the dismissed bucket.

timlevett: True, hence the :information_source: `fa-icon`, like i stated above, that could be better. More obvious or something.

keirserrie:
Thoughts--

- "Archived" may be a better name than "dismissed," as it seems more likely that stale content would be auto-archived.

- "New" sets an expectation of, well, newness. Old notifications wouldn't be there. Logically, they're now in the archived section.

- Users *probably* won't want to move an expired notification back to the new bucket. Extreme edge case.

- Assuming this is an edge case, rather than providing more info w/the rollover and icon, we can launch with a blank space there. Don't answer the unasked question.

timlevett:
I thought it may also be nice to have a count next to the tab title (e.g.: New (2)   |   Archived (102) )

keirserrie:
Agree

timlevett:
@keirserrie : What are your thoughts on having some icon in place of the `x` in new if its not dismiss-able. Sort of like the `i` for dismissed.  It could be another tooltip to tell people they have to take the action in the notification in order to dismiss this notification.

keirserrie:
Related to my above comments, I'd err on the side of less...have this area blank if you can't dismiss. We can gauge user confusion/frustration. My guess is that there won't be advanced engagement w/this interface, so simplicity is best for usability and accessibility.

Open to debate...

PR #61, opened 2015-11-04
Link presence in the Central Repository from README.

apetro:
Documents that `uw-frame` is available in the Central Repository.

This in response to the question coming up of where one can get that Maven artifact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ In your existing Maven war project, add the following dependency:
 </dependency>
 ```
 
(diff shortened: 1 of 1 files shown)


nblair:
Update the version in the pom sample too?

timlevett:
Could just link to the search.maven deep search

apetro:
I'd rather not to touch the `version` in the `pom` sample in this changeset.  This is intended as a quick hit in response to the question coming up the other day of "Where can I get that artifact?".  The next changeset can answer the question "What version can I / ought I to get?". :smile:

timlevett:
I've seen other groups do x.y.z for the version number. Makes it obvious that we won't be maintaining the current number in the readme.

timlevett:
progress :+1:

PR #62, opened 2015-11-05
Fix issue with home going to / for angularjs-portal

timlevett:
+ `angularjs-portal` needs to not refresh the page completely when it goes "home"
+ Move `SEARCH.isWeb` to `APP_FLAGS.isWeb`. Makes more sense to me this way


apetro:
:+1:

keirserrie:
:+1: this will be nice

jhanstra:
:+1:

PR #63, opened 2015-11-05
MUMUP-2123 : template for access denied

timlevett:
Based on the writings of one @keirserrie (http://staging-2.uc.wisc.edu/my-uw/#coarse-grain-access), I give you the basic access denied page.

image

### DIFF
--- a/src/main/webapp/portal/main/partials/access-denied.html
+++ b/src/main/webapp/portal/main/partials/access-denied.html
@@ -1,0 +1,13 @@
+<frame-page app-header-title='Access denied' 
+            app-header-collapse='true'
+            app-header-image='img/terrace.jpg'>
+  <h2 class='center'>Sorry, you're not authorized to access this.</h2>
+  <p class='center'>
+    <i class='fa fa-exclamation-triangle fa-3x' style='color: #b70101;'></i>
+    <br/>
+    <h4 class='center'>If you're here by accident, head back to MyUW <a href='{{MISC_URLS.myuwHome}}'>homepage</a>.</h4>
+    <h4 class='center'>For help with authorization, contact the <a href='{{MISC_URLS.helpdeskURL}}' target='_blank'>DoIT Help Desk</a>.</h4>
+    <br/>
+  </p>
+</frame-page>
+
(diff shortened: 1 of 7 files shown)


keirserrie:
so sayeth the great bald oracle :+1:

src/main/webapp/portal/main/partials/access-denied.html
```diff
@@ -0,0 +1,13 @@
+<frame-page app-header-title='Access denied'
```
apetro: Is there an opportunity to confirm to the user what application it is that they're being denied access to, so they can try to understand whether they went somewhere they didn't mean or they are being denied access to an application they felt meant to access?

>  [If it’s an application, include the header to provide context.](http://staging-2.uc.wisc.edu/my-uw/#coarse-grain-access)

timlevett: That would be nice, might be challenging to do. I guess we could pass a variable via the route variables.

apetro: Maybe I'm misunderstanding how a frame-based application is intended to relate to this template.  Is the typical frame-based application intended to override this `access-denied.html` with something tweaked to provide the details @keirserrie mentions as nice-to-have, the name of the application, the link to the related service webpage and/or the app directory entry?

Should there be an optional `access-denied-config.json` where the frame-based application provides that data as, well, structured data, and that way they're not overriding an HTML template to get those details in?  And then this template gracefully uses whatever's provided in that config and falls back on this generic experience if it ain't got no details?

timlevett: Possibly something to make in the future, maybe make a story for this. The goal of this story is to have a general access denied experience, respecting the `keep it as simple as possible`. Maybe this will come later if the people demand it.

timlevett: Additionally, if frame apps want to create there own custom access-denied page, that is great. Not sure what a template would really give them. That seems more complex than just creating a partial with the `<frame-page>` directive.

timlevett: Oh, on the name i'm just going to use `NAMES.title`

nblair:
Hooray! :+1:

PR #64, opened 2015-11-05
ngView route-based class

### LINKED ISSUE #60: ngView animations
Greetings folks,

Before a pull request comes your way, I wanted to run this by the larger group.  We have a need to animate one and only one ( so far ) route change via ng-animate.  This is very doable, but targeting a specific route only becomes much trickier.  It **could** be simpler by appending the route as a class to ng-view.  

Thoughts?

Other suggestions to target only specific route changes ( w/o an additional directive that requires defining animations at the ng-route level )?

smargovsky:
For science!  And https://github.com/UW-Madison-DoIT/uw-frame/issues/60

Creates class on ng-view by appending slugified to the word route, such as `class="route-settings"` or `class="route-parent-child-grandchild"`

Resolves #60

### DIFF
--- a/src/main/webapp/frame.jsp
+++ b/src/main/webapp/frame.jsp
@@ -55,7 +55,7 @@
             <span class="fa fa-bars"></span>
           </div>
           <div role="main" id="region-main" class="col-xs-12" ng-class="{'col-sm-10 col-sm-offset-2' : ($storage.showSidebar && APP_FLAGS.showSidebar), 'col-sm-11 max-view' : (!($storage.showSidebar) || !APP_FLAGS.showSidebar)}">
-            <div ng-view></div>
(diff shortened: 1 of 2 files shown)


timlevett:
neat. :+1: One could also have thoughts about a couple classes, e.g.: `route-parent route-parent-child route-parent-child-grandchild`. This way you could have global stuff for the whole app, and custom stuff at each level. I know this is harder, but a thought.

smargovsky:
I planned on using attribute substring matching selectors in my css to hit that case:

```css
[class^="route-"].ng-enter { ... }

[class^="route-parent-"].ng-leave { ... }
```

but the other approach could be useful too.

timlevett:
Oh interesting, that works (TIL)

PR #65, opened 2015-11-06
Minor glitch on ng-class.

timlevett:
Minor bug fix on the route class. `ng-class` evaluates the thing provided, if you add `{{}}` it will try to evaluate twice, and the second time it'll be a string with `-` in it, then angular vomits.

Related to #64


nblair:
:+1:

PR #67, opened 2015-11-10
Change a few aria roles and add title

smargovsky:
Banner is used for site-wide elements ( red bar on top ), heading for heading section of a page ( large image + app title ), application with title ( on ng-view, specific to app ) to describe the app.  In tests, we were NOT getting the app title read out ( @jiayinjx - right? ); this fixes that, but further testing is needed.


timlevett:
:+1:

PR #68, opened 2015-11-11
Truncate circle button text after 24 characters

jhanstra:
See https://github.com/UW-Madison-DoIT/angularjs-portal/pull/351

### DIFF
--- a/src/main/webapp/portal/misc/partials/circle-button.html
+++ b/src/main/webapp/portal/misc/partials/circle-button.html
@@ -1,10 +1,8 @@
-<!--<circle-button data-href='' data-target='' data-fa-icon='' data-disabled='false' data-title=''></circle-button>-->
-
 <a ng-href="{{ cbDisabled ? '#' : href}}" target="{{cbDisabled ? '' : target}}">
   <div class="btn btn-primary rounded icon-button text-right" ng-class="{ 'btn-disabled' : cbDisabled}"> 
     <i class="fa {{faIcon}} fa-2x"></i>
   </div>
 </a>
 <p ng-show='title'>
-  {{title | truncate:24}}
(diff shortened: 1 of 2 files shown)


timlevett:
Hmmmm, I like the feature. I'm sure someone will use `circle-button` outside of the scope of a widget, where they may not want the title to be truncated, and that could be frustrating for developers. Maybe make the truncate number a variable? Then for our case in `lol` we just set it to 24?

jhanstra:
Good idea. Ask and you shall receive.

src/main/webapp/portal/misc/partials/circle-button.html
```diff
@@ -1,10 +1,8 @@
...
 <a ng-href="{{ cbDisabled ? '#' : href}}" target="{{cbDisabled ? '' : target}}">
   <div class="btn btn-primary rounded icon-button text-right" ng-class="{ 'btn-disabled' : cbDisabled}"> 
     <i class="fa {{faIcon}} fa-2x"></i>
   </div>
 </a>
 <p ng-show='title'>
-  {{title}}
-</p>
\ No newline at end of file
+  {{title | truncate:trunclen}}
```
apetro: How does the optional-ness of `trunclen` work?

timlevett: Had a similar thought

jhanstra: If `data-trunclen` is not given then it doesn't truncate and shows the full text. Tested with the list-of-links widget.

timlevett: cool

timlevett:
:+1:

PR #75, opened 2015-11-23
Less hardcoded settings

### LINKED ISSUE #73: Make settings less hardcoded
Right now we have a `/settings` page. Its mostly for setting/removing flags to enable alpha features. So you can ship a 1/2 built feature for demo purposes. `angularjs-portal` has been abusing the frame page for some time for things that are just relevant to it. We recently created a `/settings/web` to kind of fix that, but I had the thought that "this could be better".

I would like to propose we move this to a config file. This way apps can use this page as well. We could have a frame constant file that would provide the frame flags. Then give the opportunity to expand upon them in an app level config.

timlevett:
+ Removed all `/web` specific beta flags
+ Added in frame config for frame beta settings
+ Added in `APP_BETA_FEATURES` (an array) to `app-config.js` with an example flag
+ Cleaned up the markup a bit on the main page
+ Note, this only changes true/false beta flags. All other flags will still have to be hand built for now. #changeoften

### Default frame settings
image

### Example of how /web will use it
image

Resolves #73


vertein:
:+1:

apetro:
:+1:

PR #77, opened 2015-11-24
Sandy background, faster transition

timlevett:
tes


vertein:
:shipit:

keirserrie:
:+1: yaaas

jhanstra:
That's neat, :+1:

PR #80, opened 2015-12-03
Fix some odd ga stuff

timlevett:
+ Clean up when the href = `javascript:;`
+ Add log when we do an outbound link
+ Add page view for `/feature`


vertein:
:+1:

PR #81, opened 2015-12-03
Fix mobile header height

timlevett:
Before
image

after
image


vertein:
Just a heads up that you have two commits from https://github.com/UW-Madison-DoIT/uw-frame/pull/80 in here.

I thumbed that one, and will thumb this as well.  

:+1:

timlevett:
opps, lol

Issue #85, opened 2015-12-07
"Service Temporarily Unavailable" view


nblair:
As a person interacting with a frame app,
I'd like a helpful error message when parts or all of the application have server problems instead of undefined behavior
so that I'm not confused by the application when there is a problem.

Similar to the `/access-denied` route, we have need for a view to display when 500 class errors are happening.

What would be a good route name? Would you like numeric, like `/500`? Or something like `/server-error` instead?

PR #87, opened 2015-12-07
Add '/server-error' route and view

nblair:
This pull request is intended to address issue #85. It adds a route/view that could be display for 500 class errors and service issues.

Example:

screen shot 2015-12-07 at 10 20 33 am

### DIFF
--- a/src/main/webapp/portal/main/partials/server-error.html
+++ b/src/main/webapp/portal/main/partials/server-error.html
@@ -3,7 +3,7 @@
             app-header-image='img/terrace.jpg'>
   <h2 class='center'>We're sorry, but there was a problem with your request.</h2>
   <p class='center'>
-    <i class='fa fa-exclamation-triangle fa-3x' style='color: #b70101;'></i>
+    <i class='fa fa-exclamation-triangle fa-3x warning-color'></i>
     <br/>
     <h4 class='center'>You can try to head back to the MyUW <a href='{{MISC_URLS.myuwHome}}'>homepage</a>.</h4>
     <h4 class='center'>If you continue to see this error, contact the <a href='{{MISC_URLS.helpdeskURL}}' target='_blank'>DoIT Help Desk</a>.</h4>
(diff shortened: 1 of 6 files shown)


timlevett:
:+1:

nblair:
Any tests needed? I couldn't find any for `/access-denied` to model off of.

If we add an httpInterceptor that reroutes to this view, we'd want to test then.

apetro:
Cf. [style guidance wrt coarse grained access denied|http://staging-2.uc.wisc.edu/my-uw/#coarse-grain-access].  I like that this looks similar.

src/main/webapp/portal/main/partials/server-error.html
```diff
@@ -0,0 +1,13 @@
+<frame-page app-header-title='{{NAMES.title}} - Service Temporarily Unavailable' 
+            app-header-collapse='true'
+            app-header-image='img/terrace.jpg'>
+  <h2 class='center'>We're sorry, but there was a problem with your request.</h2>
+  <p class='center'>
+    <i class='fa fa-exclamation-triangle fa-3x' style='color: #b70101;'></i>
```
apetro: Hardcoded color?  Could this be an alias?  Do we need CSS classes for error messaging?

timlevett:
We currently don't have any tests for route logic as the apps overwrite the routes in most cases. Or were you saying you wanted to test the partial in some way?

src/main/webapp/portal/main/partials/server-error.html
ahoffmann-wi: Agreed. `<style>` tags should be avoided and a class used.

nblair: It's pretty much a copy of the access-denied partial in the same folder, and has the same faults. I'd be happy to define a class and apply to both.

timlevett: :+1:

paulerickson:
:+1: It's really content, not behavior in my opinion, so tests would belong in the downstream apps that decide when and why to route here.

nblair:
OK to merge by me. We'd like to see a 1.3.4 release today if possible.

PR #88, opened 2015-12-07
MUMUP-2263 Adds in root url to go back home

vertein:
Go home frame app, you're drunk.

Note for release notes
`
To upgrade add rootURL to miscURLS and give a url to go to when the crest is clicked.
If no rootURL is provided, the application will instead go to it's context root like it does currently.
Suggested : 'rootURL' : '/web'
`

### DIFF
--- a/src/main/webapp/portal/main/partials/header.html
+++ b/src/main/webapp/portal/main/partials/header.html
@@ -9,7 +9,7 @@
                  <notification-bell mode='mobile-bell'></notification-bell>
              </button>
              <div class="header-mobile">
-               <h3><a aria-label="{{NAMES.ariaLabelTitle}}" class="header-text" href="./" id="myuw-header"><img ng-src="{{portal.theme.crest}}" class="hidden-xs" alt="{{::headerCtrl.crestalt}}"><span>{{portal.theme.title}}</span><span class="beta-logo">{{NAMES.sublogo}}</span></a></h3>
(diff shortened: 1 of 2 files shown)


apetro:
> If no rootURL is provided, the application will instead go to it's context root like it does currently.

Hurray for backwards compatibility!

How does this defaulting work?  Because in `href="{{MISC_URLS.rootURL}}"` , `{{MISC_URLS.rootURL}}` evaluates to `null` and then href = null is interpreted as href = `./`?

src/main/webapp/portal/main/partials/header.html
```diff
@@ -9,7 +9,7 @@
                  <notification-bell mode='mobile-bell'></notification-bell>
              </button>
              <div class="header-mobile">
-               <h3><a aria-label="{{NAMES.ariaLabelTitle}}" class="header-text" href="./" id="myuw-header"><img ng-src="{{portal.theme.crest}}" class="hidden-xs" alt="{{::headerCtrl.crestalt}}"><span>{{portal.theme.title}}</span><span class="beta-logo">{{NAMES.sublogo}}</span></a></h3>
+               <h3><a aria-label="{{NAMES.ariaLabelTitle}}" class="header-text" href="{{MISC_URLS.rootURL}}" target="_self" id="myuw-header"><img ng-src="{{portal.theme.crest}}" class="hidden-xs" alt="{{::headerCtrl.crestalt}}"><span>{{portal.theme.title}}</span><span class="beta-logo">{{NAMES.sublogo}}</span></a></h3>
```
timlevett: Use `ng-href` here instead?

vertein:
@apetro, yep.  evaluates to `href` which is interpreted as `href="./"` (apparently in chrome and firefox at least)

vertein:
ng-href is cool, I'll update to that.

vertein:
updated to ng-href

timlevett:
:+1:

apetro:
What if `/web` were to be the default value, such that frame-based app developers can simply do nothing, not configure `rootURL`, and thereby adopt the recommended behavior of the upper left crest?

vertein:
@apetro good thoughts!

PR #90, opened 2015-12-08
Issue 89 : cleanup NAMES.title to be only app name

### LINKED ISSUE #89: Changing the constant NAMES.title results in changing the placeholder in the search bar
Here's the story version:

    As a developer
    I want to change the constant `NAMES.title`
    so that views like 'access-denied' and 'server-error' show my app's name.

If I change `NAMES.title` to "Staff Time and Reporting, the placeholder text in the search bar in the top nav changes to "Search Staff Time and Reporting" which I'm guessing is undesirable.

Would it seem sensible at present to change the placeholder text in the search bar to just a fixed 'Search MyUW' rather than being parameter-izable?

timlevett:
Resolves #89 
+ Moved most theme naming to the theme constant
+ Cleaned up header, search
+ `NAMES.title` is now the app name
+ Removed not needed NAMES constaints

### DIFF
--- a/src/main/webapp/js/app-config.js
+++ b/src/main/webapp/js/app-config.js
@@ -18,11 +18,7 @@ define(['angular'], function(angular) {
             'groupURL' : null
         })
         .constant('NAMES', {
-            'title' : 'MyUW',
-            'ariaLabelTitle' : 'My U W',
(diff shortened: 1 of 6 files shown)


jhanstra:
:+1:

PR #91, opened 2015-12-08
Add in my-app module to preprocessors for coveralls coverage reports

timlevett:
This is more for downstream applications that want to use this karma file.


vertein:
:+1:

PR #93, opened 2015-12-10
MUMUP-2286: Search field populated on search results page, and nowhere else

jhanstra:
Goes with https://github.com/UW-Madison-DoIT/angularjs-portal/pull/358.

Now the search field shows the thing you're searching for when you're on that page, doesn't everywhere else.

screen recording 2015-12-10 at 01 01 pm

### DIFF
--- a/src/main/webapp/portal/search/controllers.js
+++ b/src/main/webapp/portal/search/controllers.js
@@ -3,14 +3,13 @@
 define(['angular'], function(angular) {
 
   var app = angular.module('portal.search.controllers', []);
-  app.controller('PortalSearchController', [ 'miscService', '$location', '$scope', '$localStorage','SEARCH', 'APP_FLAGS', function(miscService, $location, $scope, $localStorage, SEARCH, APP_FLAGS) {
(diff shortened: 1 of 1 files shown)


src/main/webapp/portal/search/controllers.js
```diff
@@ -33,18 +31,23 @@ define(['angular'], function(angular) {
...
+              //frame app redirect
+              window.location = SEARCH.searchURL + $scope.initialFilter;
           }
-        };
+        }
+      };
+        
+      $scope.$on('$locationChangeStart', function(event) {
+        if ($location.url().indexOf('search') == -1) {
+          $scope.initialFilter = '';
```
vertein: Seems dangerous if we ever change urls, but I guess we're in a lot of trouble when we start changing any of our urls around.

vertein:
:+1:

timlevett:
Seems legit. :+1:

src/main/webapp/portal/search/controllers.js
timlevett: Future us can worry about that ;)

PR #98, opened 2015-12-15
MUMUP-2296 : Update footer

timlevett:
I know we said we didn't want to do a constant with an array, but I did it anyway. It was too easy to pass up.
+ Removed `whatsNewURL` as it was only used by the footer (and for frame we set it to null anyway)
+ Added `FOOTER_URLS` array constant. Pretty simple url, target, title
+ For frame apps I left it like it is today, just feedback and help.

http://goo.gl/ziSxki

### DIFF
--- a/docs/js/app-config.js
+++ b/docs/js/app-config.js
@@ -35,7 +35,7 @@ define(['angular'], function(angular) {
             'logoutURL' : 'https://github.com/UW-Madison-DoIT/uw-frame',
         })
         .constant('FOOTER_URLS', [
-          { "url" : "https://kb.wisc.edu/helpdesk",
+          { "url" : "/web/static/myuw-help",
             "target" : "_blank",
             "title" : "Help"
           },
--- a/docs/js/app-config.js
+++ b/docs/js/app-config.js
@@ -33,6 +33,16 @@ define(['angular'], function(angular) {
             'rootURL' : '/uw-frame/',
             'logoutURL' : 'https://github.com/UW-Madison-DoIT/uw-frame',
         })
+        .constant('FOOTER_URLS', [
+          { "url" : "https://kb.wisc.edu/helpdesk",
+            "target" : "_blank",
+            "title" : "Help"
+          },
+          { "url" : "https://my.wisc.edu/portal/p/feedback",
+            "target" : "_blank",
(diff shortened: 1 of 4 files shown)


docs/js/app-config.js
```diff
@@ -27,12 +27,23 @@ define(['angular'], function(angular) {
...
+            'back2ClassicURL' : null,
             'feedbackURL' : 'https://my.wisc.edu/portal/p/feedback',
             'helpdeskURL' : 'https://kb.wisc.edu/helpdesk/',
             'myuwHome' : 'https://my.wisc.edu',
             'rootURL' : '/uw-frame/',
             'logoutURL' : 'https://github.com/UW-Madison-DoIT/uw-frame',
         })
+        .constant('FOOTER_URLS', [
+          { "url" : "/web/static/myuw-help",
+            "target" : "_blank",
```
apetro: Since it's in the portal, should not be `target=_blank` ?

apetro:
:+1: Always be refactoring. Also :heart: the space characters in place of `\t`s.

docs/js/app-config.js
timlevett: I would think so. I would be frustrated if I went to read help on a page i was working on and the help took over that page.

apetro:
Feasibility of smoothing the upgrade path so that frame-based app developers need change less in their `app-config` ? This looks like it requires an update to app-config in upgrading to a `uw-frame` with this change, which feels a little like an "API change" in semver parlance.  

Feasibility of some kind of fall-back-on-portal-defaults behavior where in the (typical!) case where a frame-based app wants the recommended footer links, it just gets them?

Thought experiment: MyUW adds a GitHub link, and wants that link in the footer. It would be preferable to be able to add that one day in a MyUW release and have the nothing-special-to-add-to-footer frame-based apps just get the update, immediately, without having to cut releases, and without having to convince frame-based app developers to make the same change to their app-config .  

Not a huge deal, but maybe a continued opportunity to think through how to smooth the ongoing upgrade and maintenance path.

vertein:
:+1:

timlevett:
Chatted with @apetro offline, he's going to move that suggestion to a gh-issue

PR #101, opened 2015-12-15
Fix up npm scripts

paulerickson:
There were some issues in package.json preventing `npm run static` from working out of the box, which this PR should resolve:
1.  bower and http-server dependencies are declared
2.  `npm run build-static` runs before `npm run static`
3.  ';' is replaced with '&&' in shell scripts so that non-zero error codes are not ignored (i.e., if you can't go into a directory, report failure, and _don't_ start deleting things or running http-server)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,29 +2,38 @@
   "name": "web",
   "version": "1.0.0",
   "description": "angular my-uw",
-  "main": "index.js",
   "scripts": {
-    "test": "cd uw-frame-components; bower install; cd ..; karma start uw-frame-components/karma.conf.js  --single-run; cd uw-frame-components; rm -rf bower_components",
-    "build-all": "npm run build-java; npm run build-static",
-    "build-docs" : "cd docs; ./build.sh;",
-    "build-java": "cd uw-frame-java; mvn clean package",
-    "build-static": "cd uw-frame-static; ./build.sh",
-    "jetty": "cd uw-frame-java; mvn jetty:run",
-    "static": "cd uw-frame-static/target; http-server",
-    "docs" : "cd docs/target; http-server",
-    "clean": "rm -rf uw-frame-components/bower_components; cd uw-frame-java; mvn clean; cd ..; rm -rf uw-frame-static/target"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,29 +2,38 @@
   "name": "web",
   "version": "1.0.0",
   "description": "angular my-uw",
-  "main": "index.js",
```
paulerickson: also, this didn't exist so I took it out

package.json
```diff
@@ -2,29 +2,38 @@
...
-    "test": "cd uw-frame-components; bower install; cd ..; karma start uw-frame-components/karma.conf.js  --single-run; cd uw-frame-components; rm -rf bower_components",
-    "build-all": "npm run build-java; npm run build-static",
-    "build-docs" : "cd docs; ./build.sh;",
-    "build-java": "cd uw-frame-java; mvn clean package",
-    "build-static": "cd uw-frame-static; ./build.sh",
-    "jetty": "cd uw-frame-java; mvn jetty:run",
-    "static": "cd uw-frame-static/target; http-server",
-    "docs" : "cd docs/target; http-server",
-    "clean": "rm -rf uw-frame-components/bower_components; cd uw-frame-java; mvn clean; cd ..; rm -rf uw-frame-static/target"
+    "pretest": "cd uw-frame-components && bower --config.interactive=false install",
```
timlevett: does pretest auto run if you run `npm run test`?

timlevett:
:+1: love the cleanup! thanks

package.json
paulerickson: Oh yeah, I meant to call this out - it's a naming convention that if you run script `foo` the runner always looks for a `prefoo` to run beforehand and a `postfoo` to run afterward, which gives you a rudimentary dependency scheme.  Since they'll each run in their own thread, we get to drop the `cd ..` bits

timlevett: Neato :+1:

package.json
timlevett: Prob a copy/paste from a template or something

Issue #102, opened 2015-12-16
Cleanup static loading and html5mode


timlevett:
We need to do a bit of cleanup on static uw-frame
+ If you attempt to hit a child `URL`, say `/settings`, you get a 404
+ The `static.html` page is not imported, looking at `index.html` we still have the `jsp:include`. Doubt that is going to work on a static page :)

timlevett:
Maybe use superStatic instead: https://www.npmjs.com/package/superstatic

PR #103, opened 2015-12-16
Add in serverside rewrites, change to superstatic

timlevett:
resolves part of #102 for static content pages

### DIFF
--- a/uw-frame-static/superstatic.json
+++ b/uw-frame-static/superstatic.json
@@ -1,0 +1,10 @@
+{
+  "rewrites" : [
+    {"source":"/settings", "destination" : "index.html"},
+    {"source":"/notifications", "destination" : "index.html"},
+    {"source":"/features", "destination" : "index.html"},
+    {"source":"/about", "destination" : "index.html"},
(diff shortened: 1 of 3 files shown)


uw-frame-static/superstatic.json
```diff
@@ -0,0 +1,10 @@
+{
+  "rewrites" : [
+    {"source":"/settings", "destination" : "index.html"},
+    {"source":"/notifications", "destination" : "index.html"},
+    {"source":"/features", "destination" : "index.html"},
+    {"source":"/about", "destination" : "index.html"},
+    {"source":"/server-error", "destination" : "index.html"},
+    {"source":"/access-denied", "destination" : "index.html"}
+  ]
+}
```
paulerickson: :+1:
I think this is intriguing because it seems like it wouldn't be too hard to pull the rewrites and the routes from the same js/json, and prevent them getting out of sync.

timlevett: True. In the end each app is going to define there own routes, but it is handy to have the defaults.

PR #105, opened 2015-12-17
Fix loading state for static apps via html5 import feature

### LINKED ISSUE #102: Cleanup static loading and html5mode

timlevett:
Resolves #102 for the loading state part of it

### DIFF
--- a/uw-frame-static/index.html
+++ b/uw-frame-static/index.html
@@ -15,11 +15,20 @@
   <link ng-href="css/themes/uw-{{portal.theme.name}}.css" rel="stylesheet" type="text/css"/>
   <link href="my-app/my-app.css" rel="stylesheet" type="text/css"/>
   <link rel="shortcut icon" href="bower_components/uw-ui-toolkit/dist/img/favicon.ico" type="image/x-icon"/>
+  <link rel="import" href="static.html" />
 </head>
 
 <body>
 
-  <jsp:include page="/static.html" />
+  <div class='loading'></div>
+  <script type="text/javascript">
+    var content = document.querySelector('link[rel="import"]').import;
+    if(content) {
+      var el = content.querySelector('body');
+      document.body.appendChild(el.cloneNode(true));
+    }
+  </script>
+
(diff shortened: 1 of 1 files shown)


uw-frame-static/index.html
```diff
@@ -15,11 +15,20 @@
   <link ng-href="css/themes/uw-{{portal.theme.name}}.css" rel="stylesheet" type="text/css"/>
   <link href="my-app/my-app.css" rel="stylesheet" type="text/css"/>
   <link rel="shortcut icon" href="bower_components/uw-ui-toolkit/dist/img/favicon.ico" type="image/x-icon"/>
+  <link rel="import" href="static.html" />
```
timlevett: Followed tutorial http://www.html5rocks.com/en/tutorials/webcomponents/imports/

apetro: What browsers support this? Is [this](https://developer.mozilla.org/en-US/docs/Web/Web_Components/HTML_Imports)

> Firefox will not ship HTML Imports

worrying?

timlevett: That's why I check for `truthy` content. This will work with most, and browsers that don't support it just will load like it does today.

timlevett: We could add in https://github.com/webcomponents/webcomponentsjs if we wanted to support all browsers, but by that time the loading splash will be worthless. I guess we could just copy/paste in the `static.html` into this index.html page. Was just trying to avoid dup code.

timlevett: ooo, i may try to do this: https://www.npmjs.com/package/grunt-processhtml

timlevett: See https://github.com/UW-Madison-DoIT/uw-frame/pull/106 for different approach during build time.

PR #106, opened 2015-12-17
static.html injection at build time of static

### LINKED ISSUE #102: Cleanup static loading and html5mode

timlevett:
Resolves #102 , replaces #105. Basically the same thing, this just happens at build time.

See https://www.npmjs.com/package/grunt-processhtml for details on the `processhtml` task.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,17 @@
+module.exports = function(grunt) {
+
+  grunt.initConfig({
+    pkg: grunt.file.readJSON('package.json'),
+    processhtml: {
+      dist: {
+        files: {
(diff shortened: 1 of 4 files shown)


apetro:
:+1: . Also love the potential in Grunt processing that this is exploring.

timlevett:
Agreed @apetro . Might move to a full on grunt build, but we will see.

PR #109, opened 2015-12-18
Version documentation

timlevett:
+ Remove `html5mode` for documentation. gh-pages doesn't support server side rewrites and having a dynamic base is hard.
+ Add in some files for proper redirects when they hit root of `gh-pages` (redirects to latest)


vertein:
:+1:

PR #112, opened 2015-12-22
Add in beta settings link in user menu if they have that flag set in …

timlevett:
…beta settings

http://goo.gl/ei2CX5


jhanstra:
:+1:

PR #115, opened 2015-12-28
MUMUP-2298 : Add to home controller

timlevett:
Adds 2 variables to the `app-header` directive. 
+ Add to home flag
+ fname

Adds 3 things to `app-config.js`:
+ `NAMES.fname` - to be used if your app just have one marketplace entry and you want to use the add to home directive
+ `MISC_URLS.addToHomeURLS.layoutURL` - to be used to get the layout of the user to see if the user has this fname in there layout. Could be better if the server had a "has in layout" API call.
+ `MISC_URLS.addToHomeURLS.addToHomeActionURL` : the rest call to trigger when one clicks the button


vertein:
:+1:

timlevett:
after trying to use this in angularjs-portal, i think i found a bug if you are using the 2 way binding version. I may rearrange some things.

timlevett:
alright, that should be better now.

vertein:
Cool. :+1:

PR #117, opened 2016-01-05
MUMUP-2202: Update home page search bar, make mobile search look better, improve crest and title

jhanstra:
Before: 
image

After:
image

For now we can't really update the placeholder text, because our search results don't show people or web results yet, so I left the text as is. @keirserrie let me know if you think something else would be better for the text in the meantime.

You'll notice the crest is a bit larger and moved up and left, so it overflows on both sides. I think this looks better, but open to opinions.

Also, fixed mobile so it looks better:

Before:
image

image

After: 
image

image

### DIFF
--- a/uw-frame-components/css/buckyless/header.less
+++ b/uw-frame-components/css/buckyless/header.less
@@ -118,7 +118,7 @@
   }
 }
 .search-mobile {
-  background-color:#bbb;
+  background-color:#333;
   padding:12px;
   
 }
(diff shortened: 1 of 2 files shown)


vertein:
1. :+1:

vertein:
2. I like the server error in your screen shot
3. the mobile thing makes me feel better

timlevett:
Well done, :+1:

keirserrie:
Looks great! Nitpicks when possible: reduce space btw "MyUW" and searchbar and also line up right margins :+1:

jhanstra:
@keirserrie good idea wrt space between, what do you mean by right margins though?

keirserrie:
Sorry, this should make it clearer--

screenshot 2016-01-06 14 15 16

vertein:
I like pictures

PR #123, opened 2016-01-12
Fix bower install so that it can be ran a root, if one was so inclined

timlevett:
Our docker container build slaves run as root. This is apparently a problem for `bower install`


jhanstra:
Interesting :+1:

PR #130, opened 2016-01-22
Btn link hover state fix

smargovsky:
Edge case, to be sure :

Bootstrap provides .btn-link to style buttons as links.  This fixes their hover state.  By the way, LOVE having the My UW styles here instead of in the toolkit.


timlevett:
:+1:

jhanstra:
:+1:

PR #133, opened 2016-01-28
MUMUP-2334 : Silent login on load

timlevett:
+ During loading if a recent login is not detected (last 4h) then it does a silent login via `/portal/Login?silent=true`. 
+ The response is JSON, which we store in the `$sessionStorage.portal` object. username, lastLoginMills, and theme.
+ During a fetch to `session.json` it gets a different username that is stored in `$sessionStorage.portal.username` it will throw a redirect to the login page
+ By default the silent login on load is disabled (as most apps won't do this, but `angularjs-portal` will)
+ Documented all new `app-config.js` entries in configuration.md. Lucky enough, if you don't want to use this feature, then just don't configure any of the new entries.

### DIFF
--- a/uw-frame-components/portal/main/services.js
+++ b/uw-frame-components/portal/main/services.js
@@ -22,6 +22,7 @@ define(['angular'], function(angular) {
                 && data.data.person.userName != $sessionStorage.portal.username) {
                   console.warn("Thought they were " + $sessionStorage.portal.username +
                    " but session sent back " + data.data.person.userName +". Redirect!");
+                  delete $sessionStorage.portal;
                   miscService.redirectUser(302, "Wrong User than populated in session storage.");
               }
             }
--- a/uw-frame-components/portal/main/services.js
+++ b/uw-frame-components/portal/main/services.js
@@ -4,7 +4,7 @@ define(['angular'], function(angular) {
 
   var app = angular.module('portal.main.services', []);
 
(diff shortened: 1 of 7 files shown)


apetro:
> Lucky enough, if you don't want to use this feature, then just don't configure any of the new entries.

:heart:

uw-frame-components/portal/main/services.js
```diff
@@ -15,6 +15,16 @@ define(['angular'], function(angular) {
...
       userPromise = prom.then(
           function(data, status) { //success function
+            if(APP_FLAGS.loginOnLoad) {
+              //quick check to make sure you are who your browser says you are
+              if($sessionStorage.portal
+                && $sessionStorage.portal.username
+                && data.data.person.userName != $sessionStorage.portal.username) {
+                  console.warn("Thought they were " + $sessionStorage.portal.username +
+                   " but session sent back " + data.data.person.userName +". Redirect!");
+                  miscService.redirectUser(302, "Wrong User than populated in session storage.");
```
apetro: Is this redirect sufficient cleanup, or should more be done here to clear sessionStorage?

timlevett: True, could delete `$sessionStorage.portal` to really get it clean

timlevett: Doing.

vertein:
:+1:

paulerickson:
Sounds like a new, backwards-compatible feature, so a minor version bump might be in order?

timlevett:
Logical. will do @paulerickson

PR #136, opened 2016-01-29
MUMUP-2357 : Add in optional background of whiteness

timlevett:
+ Add in `white-background` boolean to `frame-page` directive
+ Use said background in various pages in uw-frame
+ document said feature

### Before

Predev was down during the creation of this PR, but just imagine yucky grey background

### After

http://goo.gl/Dh0RyS


vertein:
:+1:

PR #137, opened 2016-01-29
MUMUP-2360 : Fix events

timlevett:
+ Delete old version of GA
+ Use new GA in config setup
+ Change outbound links to new GA
+ Events now work.


vertein:
:+1:

PR #138, opened 2016-01-29
MUMUP-2357 : Move styling for white-page out of frame-page so we can use it with j…

timlevett:
…ust the app-header w/o using the frame-page directive. Basically for 2 way bind app-header pages. (static, exclusive)


vertein:
:+1: makes sense

Issue #99, opened 2015-12-15
Graceful upgrading and defaulting of application config


apetro:
Status quo:

0. `app-config` tends to include some configuration that's default, that isn't saying what's different about this app. That is, we're [repeating ourselves](http://c2.com/cgi/wiki?DontRepeatYourself).
1. The base `app-config.js` changes in `uw-frame` releases and those changes often need to be copied into the `app-config` of the frame-based application upon its upgrading frame versions. That is, if we [think of app-config as an API, strictly speaking, the API is changing too much too often in not-strictly-backwards-compatible ways](http://semver.org/).
2. Some user-visible angularjs-portal changes ought to routinely be reflected in frame-based apps for consistency ( ~~sidebar content changes~~, footer link changes) where the app doesn't have a reason to be different, but currently at the least this requires the app to update `uw-frame` versions, re-build, re-deploy. That is, configuration and defaulting is compile-time rather than run-time.

A better future:

0. `app-config` in the frame-based app includes only things the app has an opinion about, cases where the app is special or cases where it emphatically wants the current default even if the default were to change.
1. Changes to accepted defaults and additions of defaults in `uw-frame` "just work" in frame-based applications without frame-based applications having to change anything on upgrade of frame version.
2. Some forseeable changes to defaults are reflected in frame-based apps immediately upon change to MyUW `/web`, without touching the deployed frame-based app.

Ideas:

 * Interpretation layer between app-config and the actual runtime config, so that app config can be considered in the context of its version and mashed up with runtime MyUW services.
 * change strategies so that less of the page is provided "by the app" and more is provided "by the portal", rather than the chrome around the app content being compiled into the frame-based app.
 * split app-config and move around config.  Maybe there's a defaults file and then there's app-config and app-config is an overlay on the defaults. Maybe separate into more smaller config files so that a frame-based app only overwrites the configurations it wants to change and at least gets unvarnished defaults when it compile time overlays on `uw-frame`.
 * compile all the frame-based applications together with angularjs-portal into a single unified single-page-application.  Every MyUW release is an update and deployment of the entire MyUW front end, that is, including the front ends of all the applications "in MyUW".

nblair:
:+1:, ideally I'd like to only have to specify the options I want to change. At present, I have to have a full copy of the original app-config, and merge upstream changes in every time uw-frame has a new release.

timlevett:
[...]

PR #141, opened 2016-02-01
override app config for frame apps, only what is different

timlevett:
_Note this is a merge into `3.0.0-working-branch`._

I looked at the open issue #99 and thought this could be a decent solution for this.

So the problem is we want to be able to override `js/app-config.js`. 

This does that in this way:
+ Keep `js/app-config.js` but suggest people don't override it
+ Provide the `js/override.js` file that has one constant in it called `OVERRIDE`
+ If you want to override say `APP_FLAGS.showSearch` then you would just add a new entry in OVERRIDE config. e.g.: 
```js
.constant('OVERRIDE', {
            'APP_FLAGS' : {
              'showSearch' : true
            }
        })
```
+ If you don't need to override anything in `APP_FLAGS` then just don't include that object

The other big change (why I'm suggesting it for 3.0.0) is we move where to get constants from a constant, to `$rootScope.portal.config`.  This can be done for a lot of the constants, but not for all ($rootScope can't be attached to providers/factories). So not a complete solution, but it gets us pretty far.

### DIFF
--- a/uw-frame-components/config.js
+++ b/uw-frame-components/config.js
@@ -16,6 +16,7 @@ define([], function() {
         'angulartics-google-analytics' : "bower_components/angulartics-google-analytics/dist/angulartics-google-analytics.min",
         'app-config'    : "js/app-config",
         'frame-config'  : "js/frame-config",
+        'override'      : "js/override",
         'jquery'        : "bower_components/jquery/dist/jquery.min",
         'jquery-ui'     : "bower_components/jquery-ui/jquery-ui.min",
         'ngRoute'       : "bower_components/angular-route/angular-route.min",
--- a/uw-frame-components/js/override.js
+++ b/uw-frame-components/js/override.js
@@ -1,0 +1,14 @@
+define(['angular'], function(angular) {
+
+  /*Keep in sync with docs/mardown/configuration.md*/
+
+    var config = angular.module('override', []);
+    config
+        //see configuration.md for howto
+        .constant('OVERRIDE', {
+            
+        })
+
+    return config;
+
+});
--- a/uw-frame-components/portal/features/controllers.js
+++ b/uw-frame-components/portal/features/controllers.js
@@ -4,10 +4,10 @@ define(['angular','require'], function(angular, require) {
   var app = angular.module('portal.features.controllers', []);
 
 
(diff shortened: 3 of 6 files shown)


nblair:
Seems like a reasonable approach. Can you flesh out what this would look like in the overlays?

I imagine we'd maybe have an override.js in the product, but often not. Then we'd have an override.js in just the test/qa/prod overlays for our product?

Part of me is wondering if it could be a json file instead of our properties? Then something just loaded our properties into constants? Then we would be authoring in a more familiar syntax, rather than having to remember that specific `.constant` struct.

timlevett:
@nblair Thanks for the input, much appreciated.

Yes, in an overlay project you would just add the config you want changed in the OVERRIDE constant. 

Interesting thought on just having a JSON file. Might be possible, but may be more pain than what it is worth. Since we are replacing all the `.constant`'s from `app-config.js` with just one `.constant(OVERRIDE)` that the body is just JSON. There will be copy/paste-able documentation on this as well, so you need to remember nothing. Would that be a good compromise? Another approach could be a `$http.get` to a json file on the server. However this would slow down loading time...

smargovsky:
:+1: for the `override.js` and `.constant(OVERRIDE)`.

:-1: for more `$rootScope` - how does this improve our situation versus using `.constant`?

timlevett:
Note that we have a object `portal` on `$rootScope` that will prevent apps from conflicting with any constants (if that is what you are worried about)

smargovsky:
@timlevett - I understand the problem here re: constants, but isn't that what the `.value` provider is for ( a constant that can be intercepted by decorators, but **not** injected into configs )?  

My argument against `$rootScope` is the generic argument against global variables ( i.e. http://c2.com/cgi/wiki?GlobalVariablesAreBad )

timlevett:
Good point on the `.value`, I had forgotten about it. That is helpful, thanks.

I do agree generally global variables are "bad" and you should use caution to namespace them if they are used. However, if you are using something like a global variable, than it is a global variable ;) #opinionNotFact

timlevett:
When I get time, I'm going to revamp this with `.value` and see if its better.

paulerickson:
`<require-js-plug>`
RequireJS has this cool way to define [configuration for specific modules](http://requirejs.org/docs/api.html#config-moduleconfig), so that the config can be consolidated, without polluting global scope.
It looks like most (or all?) of the config isn't truly global, but is used only in one part or another, so this might be a good alternative to `$rootScope`, in conjunction with your override logic.

Config is defined with the module name, and then it can be referenced off of the special `module` module.  Here's a simple example of specifying an endpoint URL:
https://git.doit.wisc.edu/adi-ia/transfer-verification-app/blob/master/app/config.js
https://git.doit.wisc.edu/adi-ia/transfer-verification-app/blob/master/app/service.js

timlevett:
I'll post round 2 sometime in the future, until then...

PR #142, opened 2016-02-02
Fix hover state of inner nav

jhanstra:
Goes with https://github.com/UW-Madison-DoIT/angularjs-portal/pull/389. 

Does anyone know when or why the border-bottom was changed from 3px to 4px? I think it should be 3px but I'm not sure if this breaks something someone else did.

Before (when hovering on 'MyUW'):
image

After:
image


Doug-Reed:
:+1:

jhanstra:
Turns out this was my doing, last March. I'm fairly confident it should be 3px, but let's keep an eye on that inner nav.

timlevett:
that'll happen. :+1:

PR #143, opened 2016-02-08
Cleanup feature template to be real

timlevett:
Added in some notes to cleanup some gotchas

### DIFF
--- a/uw-frame-components/staticFeeds/features_template.json
+++ b/uw-frame-components/staticFeeds/features_template.json
@@ -16,16 +16,14 @@
       "startMonth": 7,
       "startDay": 11,
       "endYear": 2015,
-      "endMonth": 7,
+      "endMonth": 7, (0 based index)
       "endDay": 25,
       "buttonText": "Continue to MyUW"
     },
     "buckyAnnouncement" : {
       "shortTitle" : "",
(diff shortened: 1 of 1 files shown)


uw-frame-components/staticFeeds/features_template.json
```diff
@@ -16,16 +16,14 @@
       "startMonth": 7,
       "startDay": 11,
       "endYear": 2015,
-      "endMonth": 7,
+      "endMonth": 7, (0 based index)
```
apetro: How safe is this approach to adding comments to `JSON`? Googling about this [feels bleak](http://stackoverflow.com/questions/244777/can-i-use-comments-inside-a-json-file)?

jhanstra: Agreed, that seems sketchy

timlevett: This is a template for `features.json`, not the actual JSON. ;)

apetro: Ah, hadn't caught that it was a template.

:+1: 

Moving this into the documentation sounds like a good idea, but I don't think it needs to block this bit of inline documentation.

PR #144, opened 2016-02-08
Fix button hover animation for flat buttons

timlevett:
Flat buttons had the hover animation, which doesn't really make them flat. This fixes that


apetro:
:+1:

jhanstra:
:+1:

PR #146, opened 2016-02-09
Responsive marketplace styles

jhanstra:
Goes with https://github.com/UW-Madison-DoIT/angularjs-portal/pull/399

### DIFF
--- a/uw-frame-components/css/buckyless/responsive.less
+++ b/uw-frame-components/css/buckyless/responsive.less
@@ -55,15 +55,6 @@
     vertical-align:center;
     padding:7px 0px 0px 0px;
   }
-  .marketplace .action-buttons {
-    display:none;
-  }
-  .portlet-list-item {
-    min-height:90px;
-  }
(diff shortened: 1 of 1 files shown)


Doug-Reed:
:+1:

apetro:
Maybe a poor question: why are "Browse" styles here in `uw-frame` when the app directory browsing is an `angularjs-portal` feature?

timlevett:
That is a very good point. We do have less in angularjs-portal :)

jhanstra:
Yea, I did have an uneasy feeling making that change. I think they were originally there to keep all responsive styles in one place, i.e. only need one media query. But I'm thinking it's time to ditch that idea and put these styles in marketplace.less in `angularjs-portal`. Good call @apetro.

timlevett:
I don't think having 2 media queries will break the bank will it?

jhanstra:
Styles have been moved.

timlevett:
Sometimes less is more :+1:

PR #147, opened 2016-02-10
Fix anchor tags with btn classes

timlevett:
Hover for an anchor in a paragraph for a `.btn` was doing funky things. This fixes that.

### Before
http://goo.gl/aYSZVZ

### After
http://goo.gl/9QaQgr


jhanstra:
:+1:

PR #148, opened 2016-02-12
Guest mode revisited

timlevett:
Adds in a check to see if you are a guest during silent login. This should remove edge cases where user isn't loaded yet but we want to check for `$rootScope.GuestMode`


Doug-Reed:
:+1:

apetro:
Late to the :+1: party, but emphatic :+1: : this is solid, important attention to detail.

PR #149, opened 2016-02-12
MUMUP-2362: Improve handling of logo

jhanstra:
Madison:
image

River Falls:
image

System:
image

All logo images should be exactly 52px high. No restriction on width but shouldn't be more than 120px or so. I made logo image files of the form uw-(college)-52.png to make it clear that it should be used as a logo.


timlevett:
:+1:

keirserrie:
:+1: yes!!!!

timlevett:
SO glad we did this now instead of after system implementation. Future us will be thankful for this moment.

PR #150, opened 2016-02-15
Thinking this will fix the # getting added to ga page views

timlevett:


timlevett:
doing some remote testing so wanted to merge. sorry for the inpatients.

PR #153, opened 2016-02-19
MUMUP-2353: Mobile burger menu looks yucky

jhanstra:
image

In Incognito, when What's New would show:
image

With search bar in focus: 
image

I'm not a huge fan of the black background - @keirserrie, did you have any suggestions for the background color? Everything else I tried wasn't really an improvement.


timlevett:
looks neat, great progress! What's up with the big gap between the first entry and the header?

jhanstra:
@timlevett it's to leave space for the search bar. I can't think of an easy way to have the menu be closer to the top without being covered up when you open search. Mostly because Bootstrap's mobile nav classes area a mess. Any ideas?

timlevett:
Chatted f2f, decided to go with a variable that knows whether the the search is open or not. if it is, then add a class to the UL with the menu options to give it a big top padding. Should fix it.

jhanstra:
@timlevett great idea, got it working
screen recording 2016-02-19 at 04 09 pm

keirserrie:
@jhanstra - i had it mocked up on white bg, little more subtle and matches other dropdowns we use:

screenshot 2016-02-19 16 11 43

jhanstra:
Ignore the fact that it looks like there's no padding on the right side - it's because my mouse is plugged in which puts in a scrollbar. See first screenshots for how it would actually look.

jhanstra:
@keirserrie cool that does look much better. where is that mockup?

jhanstra:
Also what color will the search bar be then? Dark border?

apetro:
Consider darkening the background and making it a giant click surface for dismissing the menu.

Not-awesome mockup:

darken_me

jhanstra:
Added a white background and used @apetro's idea for a box-shadow to draw focus. Didn't make the background a click surface as I think this may be difficult to do. Maybe worth revisiting that later?

I also just realized the mockup has black links...should I change them to black @keirserrie?

Screenshot:
image

apetro:
Eh. If my tap in the box-shadow will "press through" to say launch email or Scheduling Assistant, then I think we're not ready for the box-shadow -- I think it really sets the user expectation that they can click in that space to dismiss the menu without taking action on the greyed out things.

jhanstra:
One more screenshot. Thoughts?

image

timlevett:
I say :shipit: :+1:

jhanstra:
I think generally upper right is the default location for X's? Maybe not, @keirserrie thoughts? It does look less out of place when search is not active.

vertein:
Maybe no 'X' and the close is just hitting the menu button again?

jhanstra:
@vertein that's true already, but I don't think it's super obvious to the user.

jhanstra:
:shipit: without 'x'

timlevett:
:+1:

PR #156, opened 2016-02-23
MUMUP-2406 Implement a uw-frame skin for UW Stevens Point.

jiayinjx:


uw-frame-components/css/themes/uw-stevens-point.less
```diff
@@ -0,0 +1,3 @@
+@import "common-variables.less";
+@import "uw-stevens-point-variables.less";
+@import "../angular.less";
```
timlevett: Please move `@import "../angular.less";` to line one of this file.

uw-frame-components/css/themes/uw-stevens-point-variables.less
```diff
@@ -0,0 +1,16 @@
+/* UW Stevens Point colors */
+@color1: #512698; 
+@color2: lighten(@color1, 5%);
+@color3: darken(@color1, 5%);
+@color4: #E1DCCC;
```
jhanstra: Have we confirmed these are the right colors?

jhanstra:
Looks great! Nice job @jiayinjx

timlevett:
:+1:

uw-frame-components/css/themes/uw-stevens-point-variables.less
timlevett: Only color1 was specified in the design subtask. Lets go with this and see what it looks like. I'm not sure how color4 is used.

jiayinjx:
uw-stevens-point_skin
Screenshot as need.

PR #157, opened 2016-02-23
MUMUP-2366 : Make search field updateable from another controller

timlevett:
I need to be able to update the global search from the search results controller. By adding this watch we can do just that.

### DIFF
--- a/uw-frame-components/portal/search/controllers.js
+++ b/uw-frame-components/portal/search/controllers.js
@@ -3,21 +3,29 @@
 define(['angular'], function(angular) {
 
   var app = angular.module('portal.search.controllers', []);
-  app.controller('PortalSearchController', [ 'miscService', '$location', '$scope', '$localStorage','$routeParams','SEARCH', 'APP_FLAGS', function(miscService, $location, $scope, $localStorage, $routeParams, SEARCH, APP_FLAGS) {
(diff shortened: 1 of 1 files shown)


vertein:
:+1:

Doug-Reed:
:+1: 

Very cool.

paulerickson:
Consider extracting search to a service, instead of coordinating through `$rootScope`

timlevett:
Did consider that, however since we would need to broadcast the change from one controller to another, we would need to use the digest cycle. The exact use case is:
+ search for a term
+ Click on details of the term, which whips out the search field
+ Then click back to search

In that last step the search bar controller is not reinitialized and was cleared by the details action, so we don't know that a service object would have been updated, but since we are plopping on `$rootScope` it gets picked up on the next digest.

paulerickson:
But if both controllers depend on the same service, wouldn't this work just as well without `$rootScope`?
```
Controller.$inject = ['SearchService'];
function Controller(searchService) {
    ...
    $scope.service = searchService;
    $scope.$watch('searchService.getQuery()', function(new, old) {
        $scope.someLocalState = new;
        searchService.resetQuery();
    });
}
```

paulerickson:
Also, I couldn't help noticing that 'portalSearchTerm' is not referenced anywhere else, so it feels as if a new API function is being exposed.  IMHO, it shouldn't go in a patch version snapshot unless it amends existing behavior.

timlevett:
That makes sense @paulerickson about the minor version bump, added. It is being used in angularjs-portal (linked PRs together above). Thanks for the example, i'll play around with it and see if I can get it to work for my use case.

PR #158, opened 2016-02-23
MUMUP-2366 : Change query setting/getting to a service call

timlevett:
Based on excellent feedback from @paulerickson in #157, changed a `$rootScope` digest cycle to a service watch. Seems a little odd to watch a service method, but works.


paulerickson:
:+1: :smiley: 

I agree that watching a service is a little odd, and personally I always look for a way to not use `$watch`, because it can be very expensive, but I don't know what would be a clean & easy alternative here.  Observer pattern seems like a good fit, but maybe not worth the added lines of code.

I see the "minor version bump" label, but shouldn't _uw-frame-java/pom.xml_ be updated with this change?  I may be fuzzy on the project structure, but that is the actual deliverable component, vs. _uw-frame-components/bower.json_ or _./package.json_, right?

timlevett:
Good question @paulerickson. We normally only care about the version when cutting the release. Snapshot version doesn't really matter for us and can be frustrating if it is constantly changing between releases. We use the labels to determine what the next version will be. So if there is at more than 1 minor version bump, then we do a minor release next release.

paulerickson:
I think I follow, but I would strongly caution against using master branch in this way!  A common practice is to keep the master or trunk codeline shippable or production-ready at all times, and IMO a very good practice if you want to adopt CD in the future, or have to switch between other projects using a CD model.

For this use-case, I've had a lot of success in the past with using a _separate_ persistent branch, like `development` or `unstable` with a persistent snapshot version of the same name.  With this model, `unstable` is open for ongoing changes, and is only merged to master when it's ready to ship / vend / deploy.  It's a good step toward CD, too, because you can confidently ship from `master` at any time (e.g. with every commit), while also having a safe & frictionless place to integrate work without the ceremony of versioning or risk of shipping code prematurely.

Just a suggestion!

timlevett:
Appreciate the feedback, I've been toying with the idea of adopting the git-flow module. There they have a `development` branch, `master`, etc... I've used that for other projects with much success, and sourcetree has that feature built right in. I'll take a gander at CD model. Have a good doc?

paulerickson:
I've been curious about git-flow too, but my impression so far is that it may be great for projects with lots of cross-cutting code and lots of busy developers (monoliths), but I suspect it would be heavyweight smaller repos with less thrash.

I said "CD model", but I guess I really mean "having Jenkins automatically release from a certain branch", which more-or-less depends on having that pristine branch.  It could just as well go the other way -- start in `master` _(or features branches thereof)_ and end up in a branch named `stable` or `QA' or `prod`, etc.

I would be interested how automated releases could look in a git-flow model, too, since I've never actually used it

timlevett:
The issue with releasing just on a branch is that branch can change, which _can be_ very dangerous. Its good to tag, tag says "this exact point in time". Tags are also good for rollback plans. I also agree its good to keep master stable, and I think we do that. We would never merge a PR with a :-1:.

I also really like the idea of github's release setup, https://github.com/blog/1241-deploying-at-github. Basically test in production approach. Lot of caution here tho. :)

PR #162, opened 2016-02-24
App config override round 2

### LINKED ISSUE #99: Graceful upgrading and defaulting of application config

timlevett:
This is round 2 of `override.js` based on recommendations in #141. Resolves #99

Basically we change everything in `app-config.js` to `.value` instead of `.constant`. Then we have a function that goes through the `OVERRIDE` constant and updates anything that it needs to. Arrays are appended to.

### DIFF
--- a/uw-frame-components/js/override.js
+++ b/uw-frame-components/js/override.js
@@ -6,8 +6,8 @@ define(['angular'], function(angular) {
     config
         //see configuration.md for howto
         .constant('OVERRIDE', {
-            'APP_FLAGS' : {
-              'features' : false
+            'FEATURES' : {
+              'enabled' : true
             }
         })
 
--- a/uw-frame-components/js/override.js
+++ b/uw-frame-components/js/override.js
@@ -1,0 +1,16 @@
+define(['angular'], function(angular) {
+
+  /*Keep in sync with docs/mardown/configuration.md*/
+
+    var config = angular.module('override', []);
+    config
+        //see configuration.md for howto
+        .constant('OVERRIDE', {
+            'APP_FLAGS' : {
+              'features' : false
+            }
+        })
+
+    return config;
+
+});
--- a/uw-frame-components/js/override.js
+++ b/uw-frame-components/js/override.js
@@ -6,9 +6,7 @@ define(['angular'], function(angular) {
     config
         //see configuration.md for howto
         .constant('OVERRIDE', {
-            'FEATURES' : {
-              'enabled' : true
-            }
+            
         })
 
     return config;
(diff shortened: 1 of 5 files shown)


timlevett:
Right now I'm marking as a minor version bump. Upgrade path is pretty easy, and it doesn't break anything if you still override `app-config.js`. So minor unless anyone thinks otherwise.

uw-frame-components/js/override.js
```diff
@@ -0,0 +1,14 @@
+define(['angular'], function(angular) {
+
+  /*Keep in sync with docs/mardown/configuration.md*/
```
apetro: typo in path: "markdown" rather than "mardown".

apetro:
:+1:

vertein:
:+1:

PR #163, opened 2016-02-25
MUMUP-2423 : Guest mode for user option dropdown

timlevett:
#### Before
http://goo.gl/U2Ot9r
#### After
http://goo.gl/mwft2L

_disregard the fact that no bell after, just didn't have it enabled in uw-frame_


apetro:
:+1:

PR #171, opened 2016-03-01
upgrade ui-bootstrap to 1.2.1 from 0.13.4

timlevett:
Upgrades `ui-bootstrap` to the latest version.

Upgrade path notes, change the name of all the directive usages (but not any of the classes or parameters). See https://github.com/angular-ui/bootstrap/wiki/Migration-guide-for-prefixes for details

### DIFF
--- a/uw-frame-components/portal/search/partials/search.html
+++ b/uw-frame-components/portal/search/partials/search.html
@@ -7,7 +7,7 @@
             aria-label="Search bar enter the app you are looking for"
             focus-me="showSearchFocus"
             ng-show="$storage.typeaheadSearch"
-            typeahead="title as match.title for match in filterMatches | limitTo:10"
+            uib-typeahead="title as match.title for match in filterMatches | limitTo:10"
             typeahead-on-select="onSelect($item, $model, $label)"
             typeahead-loading="portletListLoading"
             typeahead-min-length="2"
--- a/docs/markdown/widgets.md
+++ b/docs/markdown/widgets.md
@@ -28,7 +28,7 @@ Widget template content is a great way to get some content right on the homepage
       <div class="widget-frame" id="portlet-id-">
         <div class="widget-header">
           <div class="widget-info">
-            <i title="Info" class="fa fa-info-circle" tooltip="" tooltip-trigger="mouseenter" tooltip-placement="top" tooltip-popup-delay="200"></i>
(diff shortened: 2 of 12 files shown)


vertein:
major or minor bump?

uw-frame-components/portal/search/partials/search.html
```diff
@@ -7,7 +7,7 @@
             aria-label="Search bar enter the app you are looking for"
             focus-me="showSearchFocus"
             ng-show="$storage.typeaheadSearch"
-            typeahead="title as match.title for match in filterMatches | limitTo:10"
+            uib-typeahead="title as match.title for match in filterMatches | limitTo:10"
```
paulerickson: Is `typeahead` broken if you don't update it?  We're using this in STAR, so to us it would be a major/breaking change if so, or a minor/non-breaking change if not.

timlevett: I'm not sure how you are using it in star. I'd have to look at it. If you are using any ui-bootstrap directive you just have to change the name to prefix a `uib-`

paulerickson: As a _general_ rule, I say err on the side of caution, since version numbers are cheap but bugs & deprecations are not.  I.e.: better for Star to update these directives and not have to than for Star to troubleshoot an unexpected bug and uw-frame to have to deprecate a released version.

And along those lines I would just recommend again keeping accurate, semantic versions for snapshots as well as release versions!  That way it's at least possible for a downstream project to be on the cutting edge and ferret out bugs before a deprecation becomes necessary!  _(not that we would want to actually do that with Star at this time)_

paulerickson: @timlevett shouldn't the rest of these change too? (lines 11+)

We're using it in a totally similar way to [search for projects](https://git.doit.wisc.edu/adi-ia/star/blob/master/star-app/src/main/webapp/my-app/timecard/timecard.html#L120)

timlevett: Nope, according to [there latest doc](https://angular-ui.github.io/bootstrap/), all the attributes do not add in the `uib-` prefix

paulerickson: Weird.  Well, if you say we _"have to change"_ something as a dependant, then that sounds like a major version to me.

timlevett: Agreed with the weirdness.

timlevett:
It was decided that this change would be a "major change" so we are going to move this to the `3.0.0-working-branch` in https://github.com/UW-Madison-DoIT/uw-frame/pull/172

PR #173, opened 2016-03-01
MUMUP-2407: Priority Notifications

jhanstra:
image

Things to note:
- Priority notifications are a new 'mode' of notifications, not a separate directive, per discussion with @timlevett. I renamed this file from notification-bell.html to notification-partials.html to better reflect what's in it.
- Notifications now have a new boolean field 'priority'. Only one notification at a time should have priority set to true.

- I did some heroics to make the caret work on desktop, because I like that it gives context to what you're looking at. I was having trouble getting its position to be dynamic with the name length, but got it to work by including it in the notification bell div instead of attached to the priority notification bar. No caret on mobile though :/
image
image

- Here's how it works on mobile:
image
image


uw-frame-components/portal/main/partials/header.html
```diff
@@ -1,43 +1,45 @@
...
+                 <a aria-label="{{portal.theme.ariaLabelTitle}}"
+                    class="header-text" ng-href="{{MISC_URLS.rootURL}}"
+                    target="_self"
+                    id="myuw-header">
+                    <div class='header-inner'>
+                      <img ng-src="{{portal.theme.crest}}"
+                           class="hidden-xs crest"
+                           alt="{{portal.theme.crestalt}}">
+                      <span class='title'>{{portal.theme.title}}</span>
+                      <!-- <span class="beta-logo">{{portal.theme.subtitle}}</span> -->
```
apetro: Inadvertent change to disable the subtitle again?  I think we're trying to optionally use this for the system campus themes in their transitions to new MyUW?

uw-frame-components/portal/notifications/directives.js
```diff
@@ -11,16 +11,17 @@ define(['angular', 'require'], function(angular, require){
         }
     });
 
-    app.directive('notificationBell', function(){
+    app.directive('notificationPartials', function(){
```
apetro: At the risk of being pedantic: "partials" is an AngularJS implementation detail, it's not awesome to expose it in the name of the resulting directive HTML element. What if we were to go a different direction here, maybe invent an AngularJS-portal namespace, and call the directive "ap-notification", for a resulting semantic-ly-named HTML element (directive) usage looking like `<ap-notification mode="priority" class="visible-xs"></ap-notification>` ?

This seems nicely readable. What is it? An AngularJS-portal notification. Oh. What mode is it in? Priority.

apetro: At the further risk of being pedantic: renaming a directive among the `uw-frame-components`, is that an API-breaking change? Necessitating a major version bump to uw-frame 3.0? (And hey, if we're bumping to 3.0 anyway, and we like the `ap-` namespace idea, opportunity to adopt that namespacing for all our directives. :smile:.

uw-frame-components/portal/main/partials/header.html
jhanstra: Hmm, that might have come from rebasing and trying to fix a merge conflict. I'll uncomment.

jhanstra:
By the way, we should be picky about the length of the priority notification text. Anything longer than this example will cause problems on mobile with the current implementation. And making this div's height flexible to accommodate any length is really hard if not impossible with our current menu/header setup.
image

uw-frame-components/portal/notifications/directives.js
jhanstra: That seems like a good idea, although I think 'ap' might be confusing - looks too much like the word 'app'. But some kind of semantic namespacing would be good. 

Yea, I suppose changing that directive name is a breaking change - good catch. Should we stick with what it was before until 3.0.0? And at that point we can talk about namespacing directives?

timlevett: Why not `uw-`

uw-frame-components/css/buckyless/notifications.less
```diff
@@ -133,6 +133,22 @@
   font-size: 20px;
 }
 .menu-notification-link {
+  .arrow-down {
+  	width: 0; 
+  	height: 0; 
+  	border-left: 7px solid transparent;
+  	border-right: 7px solid transparent;
+  	border-top: 7px solid lighten(@color1, 20%);
+    z-index:2000;
```
timlevett: Reason for this z-index? Seems a bit dramatic ya?

uw-frame-components/portal/main/controllers.js
```diff
@@ -73,17 +73,31 @@ define(['angular','require'], function(angular, require) {
...
     this.showLogout = true;
     $scope.showSearch = false;
     $scope.showSearchFocus = false;
     $scope.APP_FLAGS = APP_FLAGS;
     $scope.MISC_URLS = MISC_URLS;
-
+    notificationsService.getAllNotifications().then(function(data) {
+      var notifications = data;
+      var priorityNotifications = [];
+      for( var i=0; i < notifications.length; i++ ) {
```
timlevett: Could use filter filter here.

```javascript
var filteredArray = filterFilter(notifications, {priority === true})
if(filteredArray && filteredArray.length > 0) {
 $scope.hasPriorityNotifications = true;
}
```

Or could add a break in the if statement. Reduces the call through the list of notifications.

uw-frame-components/portal/notifications/controllers.js
```diff
@@ -93,6 +95,19 @@ define(['angular'], function(angular) {
...
+      var notifications = data;
+      var priorityNotifications = [];
+      for( var i=0; i < notifications.length; i++ ) {
+        if (notifications[i].priority === true) {
+          $scope.priorityNotifications.push(notifications[i]);
+          $scope.hasPriorityNotifications = true;
+        };
+      };
+    }, function(data) {});
+
```
timlevett: This seems redundant, couldn't we just filter on the current list of notifications?

uw-frame-components/portal/main/controllers.js
timlevett: Hmmm, thinking about this more, I'm saddened we have to call the service twice just to get a `boolean`. We do pass the `headerCtrl` into the notification directive, should we just use that?

Also, we would only want to run this function if notifications was enabled.

uw-frame-components/portal/notifications/controllers.js
timlevett: Also, what about group filtering? More reason to use the service call in the init function. You would just want to add the 

```javascript
    if (notifications[i].priority === true) {
           $scope.priorityNotifications.push(notifications[i]);
           $scope.hasPriorityNotifications = true;
         };
```

to the successFn

uw-frame-components/portal/notifications/directives.js
timlevett: Either way, marking with major version bump unless this is undone

uw-frame-components/portal/main/controllers.js
timlevett: Lastly, we should only call "all notifications" if we don't care about groups.

timlevett:
Looks great, the under the covers needs some work. If you would like assistance with this let me know, marking this as :-1: until the service/controllers are cleaned up.

uw-frame-components/portal/notifications/directives.js
apetro: Good point that it's `frame` rather than `angularJS-portal`. I had latched onto the notification-ness of it, which is partly about making the frame-based app feel like it's part of the portal, but would need a namespace that makes sense more widely within frame apps.

`uwf-` has some nice properties in matching the name of the larger software product and *not* grabbing the whole `uw-` namespace.

jhanstra: I like `uwf-`

uw-frame-components/css/buckyless/notifications.less
jhanstra: Yea, it's necessary to place it over the app-header. Even if it's 1000-something it doesn't show. z-index goes up to 9999, so I don't think it's too big of a deal...I'm not sure if there is a best practice for dealing with z-indexes.

jhanstra:
Thanks @timlevett, I realized as I was writing it that it was not as optimized as it could be, but didn't know how to improve it. Had enough trouble just getting it to work. Would love to get your eyes on it to help clean it up.

jhanstra:
Cleaned up the controllers and services and renamed partials back to bell. Thanks for the help @timlevett! Shouldn't need the major version label anymore, correct?

uw-frame-components/css/buckyless/notifications.less
timlevett: i did notice the app-header is 2000, that also seems silly. I'm going to add an issue to fix z-indexing in general

timlevett:
Looks sweet @jhanstra . What happens with multiple priority notifications?

jhanstra:
@timlevett currently, only the first notification marked with priority: true will be visible on desktop, and all will show on mobile. We could display multiple notifications on desktop, but it would be tricky and complicated - essentially we would create a new css class for each number of notifications, pushing the top-nav down further 50px with each additional notification. Does that make sense?

Or we can enforce only one notification showing on both mobile and desktop using limitTo. I would vote for this route. Allowing for multiple priority notifications starts to look really jarring, and kind of defeats the purpose of 'priority' notifications in the first place.

vertein:
i vote only 1 spot, but maybe with multiple notifications, the message changes to "you have too many priority notifications and the MyUW team has failed you, please click this annoying message to read them all"

timlevett:
I'm with you on the one priority at a time. We had talked about the multiple message, where did UX land with that? @keirserrie or Jess? Whereever we land, we should add in a `limitTo` or the multiple message thing.

apetro:
I'd prefer special handling for multiple priority notifications, something that says you have N high priority notifications and clicking it goes to the notifications page to see 'em all. This way the user can distinguish the situation of having exactly one priority notification and having one showing and others that are not showing.

I agree that in general priority notifications should be few, but I don't anticipate being able to guarantee that any given user only has one at a time.

timlevett:
Agreed with @apetro

jhanstra:
Now with support for multiple priority notifications!
image

This should be ready to merge, take a final look and thumb if it looks good :) @timlevett @apetro @vertein @keirserrie

uw-frame-components/portal/notifications/partials/notification-bell.html
```diff
@@ -9,10 +9,17 @@
...
       </div>
       <div ng-show='directiveMode === "mobile-menu"' class='notification-badge-mobile-menu'  ng-click="headerCtrl.navbarCollapsed = true">
        <i class='fa fa-bell fa-fw'></i>  Notifications ({{countWithDismissed()}})
       </div>
     </a>
+    <!-- Priority notifications -->
+    <div class="priority-notifications" ng-if='notificationsEnabled && (directiveMode === "priority")' ng-repeat="priority in priorityNotifications | limitTo: 1">
+      <a href="{{priority.actionURL}}" alt="{{priority.actionAlt}}" ng-if='priorityNotifications.length == 1' class="notification-message">{{priority.title}}</a>
+      <p ng-if='priorityNotifications.length > 1' class="notification-message">You have {{priorityNotifications.length}} important notifications. <a href="notifications" alt="{{priority.actionAlt}}">View your notifications.</a>
+  	</div>
```
apetro: Oh noes! It's a tab character!

apetro:
:+1:

uw-frame-components/portal/notifications/controllers.js
```diff
@@ -8,18 +8,28 @@ define(['angular'], function(angular) {
...
     var successFn = function(data){
       //success state
       $scope.count = data.length;
       $scope.isEmpty = ($scope.count === 0);
       $scope.status = "You have "+ ($scope.isEmpty ? "no " : "") + "notifications";
       $scope.notifications = data;
+      $scope.priorityNotifications = filterFilter(data, {priority : true});
+      if ($scope.priorityNotifications && $scope.priorityNotifications.length > 0) {
+        $scope.hasPriorityNotifications = true;
+        $('.page-content').addClass('has-priority-nots');
```
timlevett: Just to make sure, do we want to do a `.removeClass` in an else clause here?

timlevett:
:+1:

uw-frame-components/portal/notifications/controllers.js
jhanstra: Good idea, I haven't gone through all of the test cases but it can't hurt

PR #176, opened 2016-03-02
Add in ng-cloak to the index pages, for better loading

timlevett:
Even with the loading splash, you can sometimes get a weird loading experience. This should help with that.

https://docs.angularjs.org/api/ng/directive/ngCloak


vertein:
:+1: there's a note to add `angular-csp.css` to the html file.  I didn't read much farther down the page.  Is just copying the contents of this hypothetical file the preferred way?

timlevett:
Thanks for the info @vertein . Looking into it, that is only if we are running in csp mode, which we are not.

jhanstra:
:+1: awesome, seems legit. do you notice a difference @timlevett?

timlevett:
Hard to tell on localhost, but curious what test looks like @jhanstra

timlevett:
Well tested out on slow connections with the throttling in chrome dev tools. Seemed no better. Thanks for the thumbs, but this really doesn't do much for us.

vertein:
Shucks, it seemed like it would have been nice.

PR #184, opened 2016-03-09
MUMUP-2428 : make the color red for active

timlevett:
Before: 
https://camo.githubusercontent.com/e0f05e79ac65bf73724671f745e3bb56ea8674d1/687474703a2f2f676f6f2e676c2f4e7633336430
After:
http://goo.gl/6Lviu7

Relates to https://github.com/UW-Madison-DoIT/hrs-portlets/pull/57

### DIFF
--- a/uw-frame-components/css/buckyless/inner-nav.less
+++ b/uw-frame-components/css/buckyless/inner-nav.less
@@ -5,7 +5,7 @@
   margin:0px;
 }
 .inner-nav-container .inner-nav {
-  background-color:@white;
+  background:@white;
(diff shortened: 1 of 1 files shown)


jhanstra:
:+1:

apetro:
:+1:. I love that while the TL;DR; version is "make the color red", which was a totally fine summary of this, in the details it's `@color1`, which is to say, it'll just do the right thing as themed for different campuses.

keirserrie:
:+1:  & thanks @apetro for long version

PR #186, opened 2016-03-10
MUMUP-2401 : Improve failed state

timlevett:
+ If the silent login fails, it reports the failure, but continues with the loading experience. The app should handle the error not in the init block.
+ If the group service call fails it warns and switches to the default theme
+ If the require module fails to load it replaces the loading splash with error content:
http://goo.gl/RZJnsV

### DIFF
--- a/uw-frame-components/portal/main.js
+++ b/uw-frame-components/portal/main.js
@@ -1,6 +1,7 @@
 define([
     'angular',
     'require',
+    'jquery',
     'app-config',
     'override',
     'frame-config',
@@ -258,4 +259,9 @@ define([
 
     return app;
 
+},
+function(angular, require, $){
+  //error block
+  console.error("/portal/main.js failed to load. Going to set user screen to generic error page");
+  $('#loading-splash').html('<b>An error has occured during loading, please try refreshing the page. If the issue persists please contact the helpdesk.</b>');
 });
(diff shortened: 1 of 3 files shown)


uw-frame-components/portal/main.js
```diff
@@ -1,6 +1,7 @@
 define([
     'angular',
     'require',
+    'jquery',
```
paulerickson: You could pass this in as `$` rather than pick it up from root scope.

uw-frame-components/portal/main.js
```diff
@@ -152,6 +153,10 @@ define([
                 defaultThemeGo();
               }
               loadingCompleteSequence();
+            }, function(reason){
+              console.error("We got a error back from groupURL, setting theme to default");
```
paulerickson: I'd avoid console log statements in production code!  It's not really [standards compliant](https://developer.mozilla.org/en-US/docs/Web/API/Console/log) in general, and IIRC fails in IE9 in particular.

uw-frame-components/portal/main.js
```diff
@@ -253,4 +259,9 @@ define([
 
     return app;
 
+},
+function(angular, require, $){
```
paulerickson: :+1:

uw-frame-components/portal/main.js
timlevett: Gotcha. It would be nice if this were standard, this is a rare case that you would actually want an error in the log somewhere, ya know?

PR #187, opened 2016-03-11
Make it so override.js can add new elements

timlevett:
Before we looped through every single element to replace maybe a handful, that seemed silly.  Now it just loops over the keys in the override.js constant and replaces the value of that key in the config.js.  This also adds the ability to add new key/value pars to existing constants, like `APP_FLAGS`. Something we do pretty often in `angularjs-portal`. This may increase speed of the loading screen a bit, depending on your override.js file.


apetro:
:+1:

apetro:
Worth updating `docs/markdown/configuration.md` to document that can now add new key-value pairs? That currently reads like it's just over-riding the ones explicitly listed.

timlevett:
Totally, thanks for the reminder @apetro

vertein:
:+1:

PR #189, opened 2016-03-15
Start Material docs

jhanstra:
Beginning to put together some documentation on how we'll use Material. This is a place to put any Material components we use in MyUW, quick links to their documentation, and anything worth noting about how we use them or how we develop with them.

### DIFF
--- a/docs/markdown/home.md
+++ b/docs/markdown/home.md
@@ -16,6 +16,7 @@
 + [Widgets](#/md/widgets) : All things widgets
 
 ### Styling
++ [Material](#/md/material) : Information on how we use Material Design
 + [Buttons](#/md/buttons) : Button styling
 + [Coarse Grain Access](#/md/coarse-grain-access) : How to do access controls (Access denied, etc...)
 + [Tables](#/md/tables) : How to present tables
--- a/docs/markdown/material.md
+++ b/docs/markdown/material.md
@@ -1,0 +1,28 @@
+### Material Design
+
+MyUW will soon make a transition from Bootstrap to Angular Material, as an implementation of Google's [Material Design](https://www.google.com/design/spec/material-design/introduction.html). The main reasons for doing this are:
+
++ To allow UX Designers to focus on UW-specific designs and problems rather than reinventing the wheel for simple components
++ To have a structured, unified design philosophy that works well together. The hope is that Material 'just works' for us.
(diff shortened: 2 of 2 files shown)


docs/markdown/material.md
```diff
@@ -0,0 +1,28 @@
...
+<md-button class="md-raised">Button</md-button>
+<md-button class="md-raised md-primary">Primary</md-button>
+<md-button ng-disabled="true" class="md-raised md-primary">Disabled</md-button>
+<md-button class="md-raised md-warn">Warn</md-button>
+```
++ **[Fab Buttons](https://material.angularjs.org/latest/demo/button)**: Used for links inside of widgets. 
+```
+<md-button class="md-fab md-mini md-primary" aria-label="Use Android">
+  <md-icon md-svg-src="img/icons/android.svg" style="color: greenyellow;"></md-icon>
+</md-button>
```
timlevett: Looking at the material docs, you are actually not supposed to use the FAB buttons in this way. Checkout https://www.google.com/design/spec/components/buttons-floating-action-button.html for details. Not sure what we will do about the circle buttons in widgets, maybe something like fab, but makes one rethink the circle button approach.

timlevett: Maybe we could suggest doing something like list with avatar where the avatar is the fa icon? https://www.google.com/design/spec/components/lists.html#lists-specs

docs/markdown/home.md
```diff
@@ -16,6 +16,7 @@
 + [Widgets](#/md/widgets) : All things widgets
 
 ### Styling
++ [Material](#/md/material) : Information on how we use Material Design
```
timlevett: Tweak to say "going to use" since this isn't true for the current release.

docs/markdown/material.md
```diff
@@ -0,0 +1,22 @@
+### Material Design
+
```
timlevett: Maybe add a "timeline" section. Says what the rollout plan is

3.x -> material + ui-bootstrap w/ bs3
4.x -> material only

docs/markdown/home.md
jhanstra: Good advice, updated

docs/markdown/material.md
jhanstra: I added a section, feel free to edit

PR #191, opened 2016-03-18
White page mimic column

smargovsky:
Since the white-page is mimicking a bootstrap column,
it needs to float so it can grow to the size of its children
vertically.


timlevett:
:+1:

PR #192, opened 2016-03-21
Fix dependency injection for features page

timlevett:
The $rootScope is not needed. It was removed from the param list, but not the array. This has ill effects I won't give you the details on. This PR fixes the issue.


apetro:
:+1:

vertein:
:+1:

PR #196, opened 2016-03-29
MUMMNG-2414 : remove myuw specific stuff

timlevett:
Now that angularjs-portal has a docs page we don't need this here.


vertein:
:hocho: :+1:

timlevett:
linking https://github.com/UW-Madison-DoIT/angularjs-portal/pull/440

PR #200, opened 2016-03-29
MUMUP-2474 : Cleanup widget styles for .widget-list

timlevett:
+ Removed pointer when nothing to click
+ Added ability to make anchor a block element so you don't need the p tag
+ Added ability to remove hover highlight


vertein:
Good ol' number 200. :+1:

PR #201, opened 2016-03-30
MUMMNG-2404 Turning off old my.wisc

vertein:
We are encouraging uw-frame apps to no longer show the back to MyUW old in the footer.

### DIFF
--- a/uw-frame-components/static.html
+++ b/uw-frame-components/static.html
@@ -29,6 +29,13 @@
         margin: -8px;
         font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;"
       id="loading-splash">
+        <!--Bad browser error message-->
+        <!--[if lt IE 10]>
+        <div class="bad-browser">
+          <span class="fa fa-frown-o"></span>
+          <p>Sorry, MyUW beta does not support your browser.<br/><a href="https://kb.wisc.edu/myuw/page.php?id=51345">Learn how to upgrade your browser.</a></p>
+        </div>
+        <![endif]-->
         <!--No javascript enabled-->
         <noscript>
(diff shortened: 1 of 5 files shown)


uw-frame-components/static.html
```diff
@@ -29,13 +29,6 @@
         margin: -8px;
         font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;"
       id="loading-splash">
-        <!--Bad browser error message-->
-        <!--[if lt IE 10]>
-        <div class="bad-browser">
-          <span class="fa fa-frown-o"></span>
-          <p>Sorry, MyUW beta does not support your browser.<br/><a href="https://kb.wisc.edu/myuw/page.php?id=51345">Learn how to upgrade your browser.</a><br/>Can't upgrade? <a href="http://my.wisc.edu/portal/Login?profile=default">Switch back to MyUW classic.</a></p>
-        </div>
-        <![endif]-->
```
timlevett: We still want this error message, just remove the "go back to classic MyUW" part.

vertein: It was commented out.  Was it meant to be uncommented?

timlevett: That is the way these work. Knowledge drop: https://msdn.microsoft.com/en-us/library/ms537512(v=vs.85).aspx and http://www.quirksmode.org/css/condcom.html

vertein: Sounds good.  Just wanted to make sure that was done on purpose.  Changed back.

timlevett:
Odd this is a MUMMNG but no blocker. :+1:

vertein:
I didn't even think about it.  I'll change our jira key over.

PR #204, opened 2016-03-31
MUMUP-2483 : Upgrade angular and require

timlevett:
**Warning**: Do a `npm run clean` before trying to build this as things won't work otherwise.

+ Upgrade angular to 1.5.3 (from 1.4.9)
+ Upgrade require to ~2.2.0 (from 2.1.22)

Poked around, seems good.


vertein:
:+1:

PR #207, opened 2016-04-04
MUMUP-2482 make priority a priority

timlevett:
Change sort to be priority (true first), then id desc.
http://goo.gl/QHgTCx

_Also minor cleanup, features shouldn't be overwritten by default :)_


vertein:
:+1:

Doug-Reed:
:+1:

PR #213, opened 2016-04-22
MUMUP-2487 : Way to clear dismissed ids

timlevett:
+ Adds in user settings page
+ Adds in logic to delete keys from `/storage`
+ Cleaned up the beta card
+ Cleaned up the example page
+ Removed some nothing classes (classes that were being used, but had no definition in our less)

### New Settings page
http://goo.gl/hgnJwn

http://goo.gl/A3q954

### Cleaned up example page
http://goo.gl/PKp4Ud

### Cleaned up beta flags page
http://goo.gl/UqSND0


vertein:
:+1:

PR #215, opened 2016-04-28
Default state of mascot image doesn't throw a npe anymore

timlevett:
There was an odd edge case where a npe was getting thrown in the console if you don't have a cached theme. This fixes that.


Doug-Reed:
:+1:

PR #219, opened 2016-05-13
MUMUP-2508 Screen reader does not read account menu

jiayinjx:
As a screen reader user, I would like my screen reader to let me know the account dropdown menu exists, so that I can exercise that menu.

add 'role' and 'aria-label'


timlevett:
:+1:

PR #222, opened 2016-05-31
Upgrade to angular 1.5.6

timlevett:
+ Fixes build issue
+ Upgrade patch of angular


vertein:
:+1:

apetro:
What's going on with picking up the `animate` and `messages` dependencies?

Doug-Reed:
👍

timlevett:
@apetro - angular-material imports them. I'd rather layout all the angular that we are importing so we say when things are upgrades, where angular-material uses ^

PR #229, opened 2016-06-10
Adds crest image size clarification

vertein:
Adds clarification to crest size stating that image height will be scaled to 52 px.


timlevett:
👍

PR #230, opened 2016-06-13
Clean up docs a bit, and documented a lot about mascots

timlevett:
+ Added `/md/announcements`
+ Added that to the home screen
+ Changed `/md/*` to use...
 + circles for ul/li, like a good md list should
 + the card in `<frame-page>` directive

documentation image of the documentation, woh


Doug-Reed:
:+1:

docs/markdown/announcements.md
```diff
@@ -0,0 +1,65 @@
...
+</div>
+</div>
+<div class='col-sm-6'>
+<ul>
+<li>A) height : 40px</li>
+<li>B) width : 60px</li>
+<li>C) Transparent Background</li>
+<li>D) Eye height should be 12px down, so that you can see it during the hidden state.</li>
+<li>Can be an animated gif, but animations should be limited to blinking eyes.</li>
+<li>If you do have an arm, suggest the top of it to be under the top 16px.</li>
```
vertein: Thoughts on removing this suggestion, since the default has an arm that's doesn't follow this suggestion, or maybe we could ask @keirserrie for an updated default robot with a smaller hand.

timlevett: That is a valid point. I'll remove that suggestion.

vertein:
:+1:

PR #232, opened 2016-06-14
MUMUP-2500 : and you get a notification

timlevett:
Change defaults so `/storage/` and the notifications are enabled by default. This will provide a more streamlined experience for the user.


vertein:
:+1:

PR #233, opened 2016-06-14
MUMUP-2577 Add UWRF Freddy Falcon mascot

jiayinjx:
Add custom mascot for UWRF uw-frame theme.
As a UW-River Falls user, able to see the UWRF mascot announce features, so can be able to regard my.uwrf.edu as part of the friendly UWRF brand.


timlevett:
:+1:

PR #234, opened 2016-06-17
MUMUP-2588 Lnks reset-my-layout from usr settings

vertein:
Let's users reset their layout via the settings page.  Even uses the reset zone.

OH NO!!  I deleted all my favorites!!  Better use the reset button.

resetlayout

That's better
image

### DIFF
--- a/uw-frame-components/portal/settings/partials/user-settings.html
+++ b/uw-frame-components/portal/settings/partials/user-settings.html
@@ -15,6 +15,9 @@
         <div ng-if='NOTIFICATION.enabled'>
           <button class="btn btn-flat" ng-click="resetKey(KV_KEYS.DISMISSED_NOTIFICATION_IDS,'DISMISSED_NOTIFICATION_IDS')">Reset dismissed notifications <i class='fa fa-spin fa-circle-o' ng-if='DISMISSED_NOTIFICATION_IDS'></i></button>
         </div>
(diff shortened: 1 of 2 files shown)


uw-frame-components/portal/settings/partials/user-settings.html
```diff
@@ -15,6 +15,9 @@
         <div ng-if='NOTIFICATION.enabled'>
           <button class="btn btn-flat" ng-click="resetKey(KV_KEYS.DISMISSED_NOTIFICATION_IDS,'DISMISSED_NOTIFICATION_IDS')">Reset dismissed notifications <i class='fa fa-spin fa-circle-o' ng-if='DISMISSED_NOTIFICATION_IDS'></i></button>
         </div>
+        <div ng-if='MISC_URLS.resetLayoutURL'>
```
apetro: I ❤️  the graceful degradation (no `resetLayoutURL`? No worries.) and that the URL config *is* the config for enabling or disabling the feature (rather than having to coordinate a parallel `.enabled` flag). I feel like we're iteratively getting better and better at this kind of grace, and it's wonderful.

timlevett:
👍 for this change. One could suggest in addition to this neat feature is adding an ajax call to skip the 2 other screens but this gets the job done.

vertein:
I'd like to streamline the entire spring web flow thing called reset-my-layout.

PR #235, opened 2016-06-17
Added extra mascot measurements

jimhelwig:
This makes it a bit more clear what portion of the image is visible and hidden and where the eyes should be.


timlevett:
:+1: :x: :100:

PR #236, opened 2016-06-20
MUMUP-2396 : upgrade ga plugin for angulartics so we get rid of the ga issue

timlevett:
Since https://github.com/angulartics/angulartics-google-analytics/releases/tag/0.2.0 was tagged, we can now move forward with this.


vertein:
:+1:

Doug-Reed:
:+1:

PR #238, opened 2016-06-22
new mascot

timlevett:
Its still pretty long, 188px, but getting better... 

image

There is a point in which Freddy is hiding behind the search bar

image


keirserrie:
Progress...I suggested an alternative to them, we'll see what they say--

screenshot 2016-06-22 14 23 30

PR #242, opened 2016-06-24
Inject angular material

timlevett:
+ Add in angular material 1.0.9
+ Create a theme for uw-madison (obviously needs theme per portal theme)
+ Fixed minor style issue with padding in the search (style from material)
+ Added cards to the example page


apetro:
:+1:. I recall @Doug-Reed did some exploration of how Angular Material theming could relate to other theming; that research might become relevant if we shake loose to move forward with Material on ng1.

vertein:
:+1:

PR #244, opened 2016-07-05
Material theme to frame theme

timlevett:
+ Modifies the `THEME` structure to have an optional `materialTheme` with 3 children objects `primary`, `accent` and `warn`. These will then correlate to a theme in material during theme selection. Important thing. If the `materialTheme` is not defined, we still define a theme, but use the default pallettes.
+ Modifies the beta settings page to use `md-card`, `md-switch`, and `md-button`

image

### Other considerations
+ We could define a default pallette that all schools have
+ The material implementation = many `style` tags in the head section. 16 per theme. Read more [here](https://material.angularjs.org/latest/Theming/05_under_the_hood)


vertein:
:+1:

PR #249, opened 2016-07-06
Fix hardcoded colors for the notification bell

timlevett:
before
image

after
image

also adjusted the priority notification lighten to be 10% instead of 20%.


vertein:
:+1:

PR #251, opened 2016-07-15
MUMMNG-2656 Add in a new fav icon

timlevett:


apetro:
:+1:. I love [that](https://en.wikipedia.org/wiki/Favicon) it took MS to IE 11 to support `.png` for this.

timlevett:
@apetro its more complex even than that! : http://stackoverflow.com/questions/4014823/does-a-favicon-have-to-be-32x32-or-16x16?answertab=votes#tab-top

vertein:
:+1:

PR #253, opened 2016-07-20
MUMUP-2605 Refactor mascot and popup announcement

vertein:
Allows dismissal of popups and announcements individually rather than dismissing one and then all earlier ones are dismissed with it.

Handles legacy conversion.

Allows multiple popups as well.

Little video in action:

announcement

We really haven't been writing test cases for a lot of our stuff.  I'm going to try to learn about how we do test cases and write up some documentation on it.

### DIFF
--- a/uw-frame-components/portal/features/services.js
+++ b/uw-frame-components/portal/features/services.js
@@ -9,6 +9,8 @@ define(['angular'], function(angular) {
                                         'miscService',
                                         'keyValueService',
                                         'PortalGroupService',
+                                        '$sessionStorage',
+                                        'filterFilter',
                                         'KV_KEYS',
                                         'FEATURES',
                                         function($http,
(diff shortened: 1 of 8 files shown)


timlevett:
:+1:

uw-frame-components/portal/features/services.js
```diff
@@ -56,32 +55,245 @@ define(['angular'], function(angular) {
         return featuresPromise;
       }
     };//end get features
+    
+    /*
+     * If keyValueService is Active and there exists legacy announcement storage
+     * this will convert that to the new store all seen announcement ids rather
+     * than just the latest.  Will then delete the legacy announcement
```
apetro: maybe clarify to "Will then delete the legacy announcement key-value."

vertein: clarified wrt comment

PR #258, opened 2016-07-28
During build move ga javascript to target/vendor for static and java

timlevett:
So that Google analytics doesn't get blocked. Turns out the folder name is getting filtered as well.


thevoiceofzeke:
👍

vertein:
:+1:

PR #259, opened 2016-08-02
Copy ga javascript into /vendor during static build

thevoiceofzeke:
Adds build changes from #258 to static's build.sh file so that `npm run static` no longer results in a 404 error when it looks for vendor/angulartics.ga.

### DIFF
--- a/uw-frame-static/build.sh
+++ b/uw-frame-static/build.sh
@@ -16,6 +16,8 @@ cp superstatic.json ./target
 
 pushd target
 ../../node_modules/bower/bin/bower --config.interactive=false --allow-root install
+rm -rf vendor && mkdir vendor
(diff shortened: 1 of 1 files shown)


vertein:
:+1:

timlevett:
:+1: Hmm, interesting. It should have been copied over as part of the [grunt process-html run](https://github.com/UW-Madison-DoIT/uw-frame/pull/258/files#diff-35b4a816e0441e6a375cd925af50752cR10).

thevoiceofzeke:
I think maybe the build comments were left out? I'm not familiar with process-html, but [it looks like you have to have special comments in HTML](https://www.npmjs.com/package/grunt-processhtml) in order for the grunt task to do anything.

timlevett:
Ah, that could be. Alas, this work.

PR #260, opened 2016-08-02
Materialize uw-frame's top bar and notifications

thevoiceofzeke:
Hunker down cuz this one is a whopper...

**In this PR:**
- `header.html` and `app-header.html` each received a total overhaul and are now built with angular material
- `notification-bell.html`, `username.html`, and a number of other partials were also heavily modified to suit the new style
- Stripped out bootstrap popovers in favor of material menus
- Changed uw-madison theme's primary color to #c5050c and link color to #0479a8 (in accordance with wisc.edu)
- Added link hover transitions for smoother color changes
- Slimmed down mobile display by moving search bar into the main menu
- Several LESS files slimmed down/modified/reorganized
- Made some changes to the example page, mainly to experiment with a newer feature in Angular Material [(grid lists](https://material.angularjs.org/latest/demo/gridList))
- Adjusted the `circle-button` directive to make it a slightly smaller material button
- Increased size of the app-header gear button for mobile

**Notes:**
- Some of these changes (like the new link color) will make a bit more sense when the app-header is updated to match.
- We might want to revisit the color of the priority notification bar (interested in Jess's input on this).
- You can ignore the discrepancy in numbers of notifications in the screenshots. Those were hard-coded values for testing purposes.
- You can also ignore the app-header discrepancies in screenshots. The pictures including notifications were taken before the app-header changes were merged into this branch.
- I did my best to test things so far, and I am relatively sure there won't be any major bugs, but these changes should be tested at length by others as well.

### Desktop & Mobile (no notifications)
screen shot 2016-08-02 at 12 17 01 pm
screen shot 2016-08-02 at 12 17 09 pm

### Desktop & Mobile (with priority notifications)
screen shot 2016-08-02 at 11 26 21 am
screen shot 2016-08-02 at 11 26 39 am

### Mobile (with notification bell in hamburger)
screen shot 2016-08-02 at 11 28 39 am

### Menus (desktop and mobile)
screen shot 2016-08-02 at 11 33 08 am
screen shot 2016-08-02 at 11 32 59 am

### DIFF
--- a/uw-frame-components/css/buckyless/notifications.less
+++ b/uw-frame-components/css/buckyless/notifications.less
@@ -89,7 +89,7 @@
     color: @white;
 
     &.notification-badge-empty {
-      color : lighten(@color1, 40%);
+      color : lighten(@color1, 30%);
     }
 
     /* MOBILE BELL ON TOP OF HAMBURGER */
@@ -138,7 +138,7 @@
       > .arrow-down {
         border-left: 7px solid transparent;
         border-right: 7px solid transparent;
-        border-top: 7px solid lighten(@color1, 40%);
+        border-top: 7px solid lighten(@color1, 20%);
         z-index:2000;
         position: absolute;
         top:-14px;
@@ -162,15 +162,11 @@
 
 }
 
-
 /* PRIORITY NOTIFICATIONS */
 .hidden-xs .priority-notifications {
+  height: 46px;
   width: 100%;
-  height:50px;
-  position: fixed;
-  top: 0;
-  margin:0px -15px;
-  background-color: lighten(@color1, 40%);
+  background-color: lighten(@color1, 20%);
   z-index:1050;
   display: flex;
   justify-content: center;
@@ -178,15 +174,18 @@
   .notification-message {
     font-size:15px;
   }
-
+  p {
+    margin-bottom: 0;
+  }
 }
+
 .visible-xs .priority-notifications {
   min-height:40px;
   position: relative;
   top: 0;
   margin:0px -15px;
   padding:15px 20px 12px;
-  background-color: lighten(@color1, 40%);
+  background-color: lighten(@color1, 20%);
   .notification-message {
     font-size:13px;
   }
--- a/uw-frame-components/css/buckyless/notifications.less
+++ b/uw-frame-components/css/buckyless/notifications.less
@@ -167,7 +167,7 @@
   height: 46px;
   width: 100%;
   background-color: lighten(@color1, 20%);
-  z-index:1050;
+  z-index:100;
   position: fixed;
   display: flex;
   justify-content: center;
--- a/uw-frame-components/css/buckyless/notifications.less
+++ b/uw-frame-components/css/buckyless/notifications.less
@@ -168,6 +168,7 @@
   width: 100%;
   background-color: lighten(@color1, 20%);
   z-index:1050;
+  position: fixed;
   display: flex;
   justify-content: center;
   align-items: center;
--- a/uw-frame-components/css/buckyless/notifications.less
+++ b/uw-frame-components/css/buckyless/notifications.less
@@ -1,8 +1,8 @@
 /* Notification Styling */
 .notifications {
   .notifications-list {
-    margin: 0px;
-    padding: 0px;
+    margin: 0;
+    padding: 0;
   }
   .notifications-list .notification-item {
     font-size: 1.1em;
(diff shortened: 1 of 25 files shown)


uw-frame-components/css/buckyless/notifications.less
```diff
@@ -217,24 +209,27 @@
 
 
 // New Stuff Div
+/*
+  TODO: determine whether any of this is still needed
+*/
```
timlevett: Doing a quick search I don't believe `header-opaque` is used any longer

uw-frame-components/css/themes/uw-madison-variables.less
```diff
@@ -1,9 +1,9 @@
 /* MyUW-Madison colors */
-@color1: #b70101;         // Badger Red
+@color1: #c5050c;         // Badger Red
 @color2: #990000;         // Medium Red
 @color3: #660000;         // Dark Red
 @color4: #F7F5E8;         // Cream
-@link-color: #660000;     // Dark Red
+@link-color: #0479a8;     // wisc.edu blue
```
timlevett: Did we also want to change the material colors in `js/frame-config.js`?

uw-frame-components/portal/misc/partials/circle-button.html
```diff
@@ -1,8 +1,9 @@
-<a ng-href="{{ cbDisabled ? '#' : href}}" target="{{cbDisabled ? '' : target}}">
-  <div class="btn btn-primary circle-button text-right" ng-class="{ 'btn-disabled' : cbDisabled}"> 
-    <i class="fa {{faIcon}} fa-2x"></i>
-  </div>
-</a>
+<md-button ng-href="{{ cbDisabled ? '#' : href}}"
+           target="{{cbDisabled ? '' : target}}"
+           class="md-fab md-mini md-primary"
```
timlevett: Any thought to changing this so they are not circular since it goes against material's idea of what a FAB is?

uw-frame-components/portal/search/partials/search.html
```diff
@@ -1,33 +1,37 @@
...
-    <span class="input-group-btn">
-      <button aria-label="Search button" class="btn btn-flat btn-search">
-        <i class="fa fa-search"></i>
-      </button>
-    </span>
-  </div>
+<form name="searchForm" ng-submit="submit()" ng-hide="$storage.typeaheadSearch">
+  <md-input-container class="md-block myuw-search" md-no-float layout="row" layout-align="start center" style="margin:0;">
+    <md-button class="md-icon-button" ng-click="submit()" aria-label="submit search"
+               style="padding:0;margin:0;width:36px;height:36px;">
```
timlevett: Could we throw this in a `class`?

uw-frame-components/css/themes/uw-madison-variables.less
thevoiceofzeke: That link color is actually the base color that the accent pallette is based upon, so there's no need to make any changes to the theme. It's just a matter of using the "md-accent" class on button links to ensure consistency with normal <a> tags.

uw-frame-components/portal/search/partials/search.html
thevoiceofzeke: I tried that first, and there was some issue preventing the styles from being applied when the DOM was all assembled. The problem seemed to come from the use of the `md-menu` directive. For some reason, trying to style things using classes just doesn't work when that thing is nested inside a menu. I'm still trying to figure out why.

I don't like using inline styles, but I think it is necessary temporarily.

timlevett:
I pushed this branch to predev to do some hands on testing. Looks great but noticed a few things.
+ The mascot announcement bubble is missing
+ The beta settings option is missing even if you have the flag enabled in `/web/settings`
+ The add more to home on the app-header is hard to read (the contrast is not enough)

image

+ The options drop down is placed strange

image

+ the bell seems to high in the header compared to name

image

Responsive bugs
+ At some points the myuw overlaps the search
image

+ Whats new is misaligned in mobile menu
image

Looking solid tho, great effort here!

thevoiceofzeke:
@timlevett All those issues should be fixed.

timlevett:
Looks great on predev, I say :shipit: :+1:

timlevett:
Kind of an odd one, but looks like there is a weird edge case where there is a gap between the header and the app-header

image

uw-frame-components/portal/misc/partials/circle-button.html
thevoiceofzeke: I think that's a good call. Normal (flat) icon buttons should suffice. I'll play around with it.

timlevett:
With priority notes, if you scroll down the priority notes is not sticky but the header is and it looks odd

image

PR #263, opened 2016-08-05
Improve appearance of settings pages and clean up LESS code

thevoiceofzeke:
**In this PR:**
- Cleaned up look and feel of beta settings and user settings pages
- Added some helper text to the user settings page and simplified button text
- Removed `base.less`
- Moved "helper" css classes into `helpers.less`
- Moved css related to frame content into `frame.less`
- `general.less` now has the styles you would expect (basic, low-specificity selectors)
- Created `circle-button.less` so that the /directives directory has more purpose
- Created `settings.less` to house the styles for settings pages
- Added some comments to help people know whether the .less file they're looking at is the one they're looking for
- Changed instances of "0px" to "0" in several files so my IDE will stop screaming at me every time I open them
- Alphabetized .less import statements in `angular.less` for the sake of my sanity

### Screenshots
screen shot 2016-08-05 at 12 33 36 pm
screen shot 2016-08-05 at 12 42 02 pm
screen shot 2016-08-05 at 12 33 25 pm


uw-frame-components/css/angular.less
```diff
@@ -2,19 +2,20 @@
...
-  @import "buckyless/directives/app-header.less";
   @import "buckyless/frame.less";
-  @import "buckyless/features.less";
+  @import "buckyless/general.less";
+  @import "buckyless/header.less";
+  @import "buckyless/helpers.less";
   @import "buckyless/inner-nav.less";
   @import "buckyless/links.less";
+  @import "buckyless/notifications.less";
+  @import "buckyless/responsive.less";
```
timlevett: I think responsive should be last on purpose for CSS layering reason as it overrides default behavior depending on media queries from selectors from other less files. I personally don't like this and feel like the media queries should be in the less file that has the other selectors. Does this make sense?

thevoiceofzeke: Yes, ideally the media queries should be nested in the selector they pertain to. For now I'll move it back to being the last import. That file needs to have some unused code cleaned out, and the media queries should be adjusted. I think that work should be included in a future PR that also includes creating variables for media queries. I'll add a comment to [JIRA #2646](https://jira.doit.wisc.edu/jira/browse/MUMUP-2646).

timlevett:
👍

PR #264, opened 2016-08-08
Add option for single menu items in app-header

thevoiceofzeke:
In this PR:
- Added params to app-header and frame-page that will allow displaying a single option button (without hiding it inside a dropdown menu) when true

**Notes:**
- This will enable displaying the compact/expanded toggle in the app header in angularjs-portal
- This change shouldn't affect any existing frame apps -- you have to know about the parameter to use it
- Small screens unaffected (will display dropdown menu with gear icon regardless of whether single-option flag is true or false)
- I'm not super excited about going this route, and I view it as a short-term solution. We should still revisit how we're delivering options menus
    - I tried out @timlevett's suggestion of using a JSON list of links, which we could then format and display however we want, rather than the current approach of allowing people to write whatever markup they want for their options menu. I did this by creating a new config constant (APP_OPTIONS_MENU) that had an array of options, each with a `url` and `name` attribute. This worked great, but it does not support the kind of markup necessary for a toggle switch (which requires a controller and some elaborate HTML markup). I was exploring ways to support both the JSON list *and* custom markup, but it turned out to be a pretty deep rabbit hole. I think it would be better saved for a future story.

### Screenshot
screen shot 2016-08-08 at 12 03 27 pm

### DIFF
--- a/uw-frame-components/portal/misc/partials/app-header.html
+++ b/uw-frame-components/portal/misc/partials/app-header.html
@@ -1,7 +1,8 @@
 <md-subheader role="heading" class="app-header md-primary md-hue-2">
   <h1 flex="50" class="app-title"><i class="fa {{::icon}}"></i> {{::title}}</h1>
   <div flex="50" class="header-links">
-    <!-- Options menu -->
+
+    <!-- OPTIONS MENU WITH MULTIPLE OPTIONS -->
     <md-menu ng-if="optionTemplate && !isSingleOption">
       <md-button aria-label="Open options menu" class="md-primary link-div" ng-click="$mdOpenMenu($event)">
         <span hide-gt-sm layout="row" layout-align="center center"><i class='fa fa-gear fa-fw'></i></span>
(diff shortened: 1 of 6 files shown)


timlevett:
👍 Looks good, seems logical. If you could please update [the documentation for this directive](https://github.com/UW-Madison-DoIT/uw-frame/blob/master/docs/markdown/directives.md) that gets published in [the docs](http://uw-madison-doit.github.io/uw-frame/latest/#/md/directives) to include the new parameter

uw-frame-components/portal/misc/partials/app-header.html
```diff
@@ -12,7 +13,24 @@ <h1 flex="60" class="app-title"><i class="fa {{::icon}}"></i> {{::title}}</h1>
...
     </md-menu>
-    <!-- Add to home button -->
+
+    <!-- SINGLE-OPTION MENU -->
+    <div class="link-div" ng-if="optionTemplate && isSingleOption" hide-xs hide-sm>
+      <div ng-include="optionTemplate"></div>
+    </div>
+    <md-menu ng-if="optionTemplate && isSingleOption" hide-gt-sm>
+      <md-button aria-label="Open options menu" class="md-primary link-div" ng-click="$mdOpenMenu($event)">
+        <span layout="row" layout-align="center center"><i class='fa fa-gear fa-fw'></i></span>
```
apetro: Gear icon for decoration, such that this `<i` [would benefit](http://fontawesome.io/accessibility/) from an `aria-hidden="true"` ?

timlevett: I'd be curious to see if the gear even is attempted to be read by a screen reader because the button has an `aria-label`.

vertein:
:+1:

uw-frame-components/portal/misc/partials/app-header.html
thevoiceofzeke: Mac OS X VoiceOver utility does not attempt to read the icon, and I'd expect other screen readers to behave the same because of the `aria-label` that @timlevett mentioned.

timlevett:
:+1: Documentation looks great!

uw-frame-components/portal/misc/partials/app-header.html
```diff
@@ -1,8 +1,9 @@
 <md-subheader role="heading" class="app-header md-primary md-hue-2">
-  <h1 flex="60" class="app-title"><i class="fa {{::icon}}"></i> {{::title}}</h1>
-  <div flex="40" class="header-links">
-    <!-- Options menu -->
-    <md-menu ng-if="optionTemplate">
+  <h1 flex="50" class="app-title"><i class="fa {{::icon}}"></i> {{::title}}</h1>
```
apetro: I suspect this icon ought to be `aria-hidden`; proposed in #265 .

PR #265, opened 2016-08-09
aria-hide the app-header app-title decoration icon.

apetro:
I suspect (but don't know) that this icon ought to be `aria-hidden` because for screen reader purposes it's decoration redundant with `{{::title}}`.


thevoiceofzeke:
Can't hurt. ¯\_(ツ)_/¯

PR #269, opened 2016-08-15
Make the config.js extendable

timlevett:
... by injecting things in my-app/app-config.js

This way you can add in a paths or shims into your application, but not rewrite the whole thing. See the changes to /docs for an example.


uw-frame-components/config.js
```diff
@@ -1,51 +1,71 @@
 'use strict';
-define([], function() {
-  return {
+define(['./my-app/app-config'], function(myAppConfig) {
```
paulerickson: For reasons I don't fully grok, I think you need to add the ".js" suffix in situations like this where the runtime is not fully bootstrapped.

timlevett: Nice catch @paulerickson! Fixed, which also fixed the test run.

vertein:
:+1:

timlevett:
See http://uw-madison-doit.github.io/uw-frame/v2.6.0/#/md/injecting-dependencies for details

PR #271, opened 2016-08-17
Add in ways to run bower install via pom.xml or build.sh

timlevett:
With suggestion from @nblair added in some other ways to run bower install during the build process of the downstream application.


nblair:
👍

PR #275, opened 2016-08-17
Improve mascot announcements UX

thevoiceofzeke:
This is the first pass at improving the UX of mascot announcements. There is more to be done after this!

**In this PR:**
- Mascot no longer shows announcements automatically on hover
- Instead, it moves slightly and shows a tooltip ("click to see what's new")
- Announcements only shown upon clicking the mascot
- Increase space given to app-header text on mobile screens

### Demo
bucky-demo

### DIFF
--- a/uw-frame-components/js/override.js
+++ b/uw-frame-components/js/override.js
@@ -1,13 +1,15 @@
 define(['angular'], function(angular) {
 
-  /*Keep in sync with docs/mardown/configuration.md*/
+  /*Keep in sync with docs/markdown/configuration.md*/
 
     var config = angular.module('override', []);
     config
         //see configuration.md for howto
         .constant('OVERRIDE', {
-
-        })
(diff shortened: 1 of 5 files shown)


uw-frame-components/js/override.js
```diff
@@ -1,13 +1,15 @@
...
+  /*Keep in sync with docs/markdown/configuration.md*/
 
     var config = angular.module('override', []);
     config
         //see configuration.md for howto
         .constant('OVERRIDE', {
-
-        })
+          'FEATURES': {
+            'enabled': true
```
apetro: Is this a change to default behavior for frame-based apps? Deliberate?

thevoiceofzeke: Oops, I did not mean to add that change. Will revert. Thanks!

vertein:
:+1:

PR #277, opened 2016-08-18
Make circle buttons more prominent

thevoiceofzeke:
In this PR:
- circle-buttons are raised and given background color
- icons in buttons got slightly smaller
- circle-buttons use accent color (will add more color to themes like Milwaukee and Oshkosh)
- also removed the position change of launch buttons on hover

### Screenshots (new next to current)
screen shot 2016-08-18 at 11 24 29 am
screen shot 2016-08-18 at 11 23 04 am


timlevett:
:+1:

vertein:
The launch button position change always slightly annoyed me.  The use of accent color is interesting.  It might make the page seem more complete now with all the colors.  :+1:

PR #278, opened 2016-08-18
Cleanup add more to home for ff not handling <a> inside a <button> well

timlevett:
+ Weird edge case where firefox doesn't work well if there is a button with an anchor tag inside of it.
+ Also cleaned up the `:not('.btn'), :not('.md-button')` (two different selectors) vs. `:not('.btn'):not('.md-button')` (one selector saying if its in either). I learned that [here](https://css-tricks.com/almanac/selectors/n/not/)

### DIFF
--- a/uw-frame-components/css/buckyless/general.less
+++ b/uw-frame-components/css/buckyless/general.less
@@ -5,7 +5,7 @@ body {
 }
 a {
   transition: @link-hover-transition;
-  &:not(.md-button), &:not(.btn) {
+  &:not(.md-button):not(.btn) {
     text-decoration: none;
     &:hover {
       text-decoration: underline;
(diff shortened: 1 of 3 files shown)


uw-frame-components/css/buckyless/general.less
```diff
@@ -5,7 +5,7 @@ body {
 }
 a {
   transition: @link-hover-transition;
-  &:not(.md-button), &:not(.btn) {
```
thevoiceofzeke: Interesting. The result is the same, right? In the absence of `:not` this would break some styles, but because it's a `:not`, does it make any difference whether it's "and" or "or"?

thevoiceofzeke:
👍  assuming you tested to make sure the and/or change doesn't break anything :P

uw-frame-components/css/buckyless/general.less
timlevett: Yeah, it was weird. The way it breaks down is like its two different selectors.  Like 
```
a:not(.md-button){
 //some style
}

a:not(.btn) {
  //some style
}
```

timlevett: Corrected and/or statement above to clear it up a bit.

PR #279, opened 2016-08-19
Link color was too dark -- changed accent color and link color to mat…

JoyceJohnston:
…ch color1


thevoiceofzeke:
👍 

Any guidance you have about specific colors for your campus brand is very welcome! Over time, as more uw-frame and portal conform to material design, we'll be phasing out

PR #284, opened 2016-08-25
MUMUP-2643 : Prompt users to reauth

timlevett:
When a user pulls a @jimhelwig and leaves their browser open for really really really long periods of time, things get funky. Shib sessions timeouts so things looks like they are working but getting any server side content can be strange. This puts a timeout component in place (placed in the footer) that notifies the user "hey, you need to reauth".

image


jimhelwig:
I like this!

jjones770:
Will users see this in frame apps too? Just wondering if we need to warn users that any info entered that wasn't saved/submitted might be lost.

Otherwise, just some small tweaks to the wording:

Your Login Session Expired

Please click below to log back in, or close your window to log out.

timlevett:
@jjones770 
> will other frame apps see this? 
If they enable this feature. Maybe we could make this text changeable so they could say something about lose of data?

I can make the text changes, thanks for reviewing!

PR #290, opened 2016-08-30
Fix session vs session issue

timlevett:
Weird thing where git couldn't figure out session.json is different than Session.json


vertein:
:+1:

PR #281, opened 2016-08-24
Adjust appearance of priority notifications with buttons

thevoiceofzeke:
**In this PR:**
- Included Doug's work from the [MUMUP-2678 branch](https://github.com/UW-Madison-DoIT/uw-frame/tree/MUMUP-2678)
    - Added `actionButtons` attribute to sample notifications
    - `actionButtons` contains objects with the following attributes: `url`, `label`, `action`
- Added angular material attributes to ensure buttons appear in a row
- Added some CSS to reduce button size
- Changed background of priority notifications to generic light gray (previously we were lightening the theme's "color1," which was not always going to look good with white text)
- First button will be theme's primary color, last button will be the accent color
- Added black border around priority notifications to help set them apart from the rest of frame
- Added zero padding-left on `<ul>` lists to ensure consistency across browsers

### Screenshot
screen shot 2016-08-24 at 12 28 17 pm

PR #293, opened 2016-08-30
MUMUP-2712 No more oskhosh, but rather ohskosh

vertein:
Before (in predev):
image

After:
image


timlevett:
cause: https://github.com/UW-Madison-DoIT/uw-frame/pull/281/files#r76277086

:+1: for fix.

PR #294, opened 2016-08-30
Fix Oshkosh theme spelling

timlevett:
The spelling of Oshkosh was wrong in frame-config.js' THEMES. That was fixed in a PR, but then that broke where it was misspelled for less. Fixing the builds and the overall name of the less files for that theme.


timlevett:
Looks like @vertein did this in another pr, closed.

PR #295, opened 2016-09-01
Fix disappearing-header bug

### LINKED ISSUE #289: Header disappears when clicking the down-caret next to name if scroll position is not at the top
Using release 2.6.1.

If you scroll down, then click the down-caret next to your name, the log out button appears, but the header disappears.

This behavior does not occur if you are scrolled to the top.

Example:

frame-logout

thevoiceofzeke:
Fix for [MUMUP-2703](https://jira.doit.wisc.edu/jira/browse/MUMUP-2703). Just needed to concretely set the position of the top bar so it's not displaced when a new `fixed` element appears (the menu).

### Demo
header-fix-demo

Resolves #289


timlevett:
:+1:

vertein:
So close to making the last release. :/

:+1:

PR #297, opened 2016-09-01
Add link color to priority notifications

thevoiceofzeke:
Noticed during sprint review that links in priority notifications are black. Added some CSS to allow them to use the given theme's `link-color`.


vertein:
👍

PR #304, opened 2016-09-08
MUMUP-2711 no popups in guest mode

Doug-Reed:


vertein:
:+1:  - Also, going to do another PR with a guest mode fix for mascot features?

Doug-Reed:
@vertein 

https://github.com/UW-Madison-DoIT/uw-frame/blob/master/uw-frame-components/portal/features/controllers.js

Line 106 in the init function already has a check for guest mode. That should prevent public mascot announcement, shouldn't it? I'll test it.

uw-frame-components/portal/features/controllers.js
```diff
@@ -53,10 +53,10 @@ define(['angular','require'], function(angular, require) {
 
 
      //local functions ---------------------------------------------------------
-     
+
      var getPopups = function(){
        portalFeaturesService.getUnseenPopups().then(function(unSeenPopups) {
-         if(unSeenPopups.length !=0){
+          if(unSeenPopups.length !=0 && !$rootScope.GuestMode===true){
```
apetro: Can `!$rootScope.GuestMode===true` be idiomatically refactored to `!$rootScope.GuestMode`?

Doug-Reed: Who are you calling an idiom?

Doug-Reed:
@apetro : Refactored.

uw-frame-components/portal/features/controllers.js
```diff
@@ -56,7 +56,7 @@ define(['angular','require'], function(angular, require) {
 
      var getPopups = function(){
        portalFeaturesService.getUnseenPopups().then(function(unSeenPopups) {
```
paulerickson: "unseen" is one word, not two!

vertein: I like how @paulerickson swoops in!  I'll take blame, as I did that originally... :(  I will also point out my inconsistency with the getUnseenPopups name and the unSeenPopups variable.

uw-frame-components/portal/features/controllers.js
```diff
@@ -53,10 +53,10 @@ define(['angular','require'], function(angular, require) {
 
 
      //local functions ---------------------------------------------------------
-     
+
      var getPopups = function(){
        portalFeaturesService.getUnseenPopups().then(function(unSeenPopups) {
-         if(unSeenPopups.length !=0){
+          if(unSeenPopups.length !=0 && !$rootScope.GuestMode){
```
timlevett: Should we even call the `portalFeaturesService.getUnseenPopups()` if you are a guest? I would think this check is before this return from a service call.

PR #306, opened 2016-09-14
Create launch button directive

thevoiceofzeke:
Makes `<launch-button>` available for portal widget entities so we can provide a standard experience across widgets. Satisfies [MUMUP-2724](https://jira.doit.wisc.edu/jira/browse/MUMUP-2724)

**In this PR:**
- Added launch button directive and template
- Updated docs to describe the directive

### Screenshot
screen shot 2016-09-14 at 10 30 45 am

### DIFF
--- a/docs/markdown/directives.md
+++ b/docs/markdown/directives.md
@@ -85,13 +85,31 @@ Displays a flat, circular icon-button with a fa-icon in the middle, and a title 
 ```
 #### Params:
 * **href**: Where you want them to go
-* **target**: Open in new window
+* **target**: Open in new window, new tab, or the same window
 * **fa-icon**: The font awesome icon to use
 * **disabled**: Button disabled or not (can be a variable)
 * **title**: (optional) Title that is displayed under the circle
 * **truncLen**: (optional) Length to truncate the title
 
 <a href='#/demo' class='btn btn-flat btn-sm'>See Demo here</a>
+
+## LaunchButton
+
+Displays a launch button for portal widgets that fits their width and visual style
+
+#### Template :
+
+```html
+<launch-button 
+	data-href="" 
+	data-target="" 
+	data-title="">
+</launch-button>
+```
+#### Params:
+* **href**: Where you want them to go
+* **target**: Open in new window, new tab, or the same window
+* **title**: Launch app text (e.g. "Launch App," "Go to \[your site]," etc.
 
 ## Loading Gif
 
--- a/docs/markdown/directives.md
+++ b/docs/markdown/directives.md
@@ -93,7 +93,7 @@ Displays a flat, circular icon-button with a fa-icon in the middle, and a title 
 
 <a href='#/demo' class='btn btn-flat btn-sm'>See Demo here</a>
 
-## LaunchButton
+## Launch Button
 
 Displays a launch button for portal widgets that fits their width and visual style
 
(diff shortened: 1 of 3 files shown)


uw-frame-components/portal/misc/directives.js
```diff
@@ -252,6 +252,28 @@ define(['angular', 'require'], function(angular, require) {
...
 
+  /**
+   * Launch Button Directive
+   * Displays a button that fits the width and visual style of a widget
+   * Template: <launch-button data-href="" data-target="" data-title=""></launch-button>
+   *
+   * Params:
+   * - href: where you want them to go
+   * - target: open in new window, new tab, or same window
+   * - title: the text to be displayed
```
apetro: `title` seems potentially confuse-able with title-of-app. Would `label' be more expressive?

uw-frame-components/portal/misc/partials/launch-button.html
```diff
@@ -0,0 +1,5 @@
+<a class="launch-app-button ng-scope"
+   target="{{ target }}"
+   ng-href="{{ href }}">
```
apetro: Should this directive accept an additional parameter that provides say [an `aria-label`](https://www.w3.org/TR/WCAG20-TECHS/ARIA8.html)?  This could encourage implementing [design guidance](https://docs.google.com/document/d/1JQOe0a9XPZdTev-M7qyl_bRHn3acBSVJLkTvx57uIf4/edit) to e.g. supplement a "Launch app" visual text label with a ARIA attributes that clarify what app will be launched ("This button will launch the Time and Absence app in MyUW.").

docs/markdown/directives.md
```diff
@@ -85,14 +85,32 @@ Displays a flat, circular icon-button with a fa-icon in the middle, and a title
...
+	data-href="" 
+	data-target="" 
+	data-title="">
+</launch-button>
+```
+#### Params:
+* **href**: Where you want them to go
+* **target**: Open in new window, new tab, or the same window
+* **title**: Launch app text (e.g. "Launch App," "Go to \[your site]," etc.
+
```
timlevett: :+1: :x: :100: for providing docs

uw-frame-components/portal/misc/partials/launch-button.html
timlevett: Could even have a sensible default like [the guidance](http://uw-madison-doit.github.io/angularjs-portal/latest/#/md/widget-launch-button) suggests.

uw-frame-components/portal/misc/directives.js
thevoiceofzeke: How about `linkText`?

uw-frame-components/portal/misc/partials/launch-button.html
thevoiceofzeke: I think that's a good idea given the current guidance, but I'm also wondering if we should revisit the guidance and tweak it a bit. [Links should make sense outside of their context](http://webaim.org/techniques/hypertext/link_text), so "Launch App" isn't really a helpful link when there are 10 other "Launch App" links on the page. "Launch Time & Absence," on the other hand, is clear in any context.

I'm all for adding an optional aria-label for the time being.

jjones770:
@thevoiceofzeke Definitely see your point. I agree that for accessibility (and also context), using the name of what they are launching/opening makes sense. I think the challenge comes in with character limits. Do we maybe need to take a step back in the button label guidance and tie it to app naming guidance, or have an instituted character limit on app names? Seems like we take the road of more flexibility usually, so the former might make more sense.

PR #307, opened 2016-09-15
Add more checks for guest mode for annoucements

timlevett:
Yet another race condition check.

### DIFF
--- a/uw-frame-components/portal/features/controllers.js
+++ b/uw-frame-components/portal/features/controllers.js
@@ -110,9 +110,12 @@ define(['angular','require'], function(angular, require) {
          delete $localStorage.lastSeenFeature;
          delete $localStorage.hasSeenWelcome;
          //Mode is set to bucky or bucky_mobile to signal mascot init of controller
-         if("BUCKY" === $scope.mode || "BUCKY_MOBILE" === $scope.mode) {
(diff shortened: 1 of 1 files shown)


uw-frame-components/portal/features/controllers.js
```diff
@@ -110,9 +110,12 @@ define(['angular','require'], function(angular, require) {
          delete $localStorage.lastSeenFeature;
          delete $localStorage.hasSeenWelcome;
          //Mode is set to bucky or bucky_mobile to signal mascot init of controller
-         if("BUCKY" === $scope.mode || "BUCKY_MOBILE" === $scope.mode) {
+         if(("BUCKY" === $scope.mode || "BUCKY_MOBILE" === $scope.mode)
+            && !$rootScope.GuestMode) {
            portalFeaturesService.getUnseenAnnouncements().then(function(unseenAnnouncements) {
-             $scope.announcements = unseenAnnouncements;
+             if(!$rootScope.GuestMode){
```
apetro: So what's going on here is that `$rootScope.GuestMode` is not truthy when `portalFeaturesService.getUnseenAnnouncements()` fires, but when it returns to call the callback function, `$rootScope.GuestMode` might then be truthy so need to gate setting `$scope.announcements` on that?

Seems complicated enough to be worth an inline code comment?

PR #309, opened 2016-09-21
MUMUP-2721 : Add in dismiss to priority notifications

timlevett:
image

untitled gif

### DIFF
--- a/uw-frame-components/portal/notifications/partials/notification-bell.html
+++ b/uw-frame-components/portal/notifications/partials/notification-bell.html
@@ -44,6 +44,9 @@
           {{button.label}}
         </md-button>
       </div>
+      <div class="dismiss-priority hidden-xs">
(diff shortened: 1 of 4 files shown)


uw-frame-components/portal/notifications/partials/notification-bell.html
```diff
@@ -44,6 +44,9 @@
           {{button.label}}
         </md-button>
       </div>
+      <div class="dismiss-priority hidden-xs">
+        <md-button ng-if="priority.dismissable" ng-click="dismiss(priority, true)"><i class='fa fa-times' aria-label="Dismiss this priority notification"></i></md-button>
```
thevoiceofzeke: The aria-label attribute should be part of md-button, not the <i> element. Not sure if screen readers will pick it up as is.

timlevett: Thanks for the info @thevoiceofzeke . Will fix.

PR #320, opened 2016-09-27
Move notification items into directive

thevoiceofzeke:
For [MUMUP-2741](https://jira.doit.wisc.edu/jira/browse/MUMUP-2741)

**In this PR**:
- Moved the repeated code for notification items into a directive -- opted to go with the simplest version possible...not sure if it's worthwhile/helpful to move the buttons and "no notifications" cases into the directive
- Fixed some nit-picky syntax inconsistencies in the controller

### DIFF
--- a/uw-frame-components/portal/notifications/partials/notifications-list-item.html
+++ b/uw-frame-components/portal/notifications/partials/notifications-list-item.html
@@ -1,0 +1,23 @@
+<!-- NOTIFICATION CONTENT -->
+<a class="md-list-item-text"
+   ng-href="{{ notification.actionURL }}"
+   target="_blank"
+   aria-label="{{ notification.actionAlt }}"
+   layout="row"
+   layout-align="start center"
+   layout-xs="column"
+   layout-align-xs="center start">
+  <span><i class="fa fa-exclamation-triangle" ng-if="notification.priority"></i> {{ notification.title }}</span>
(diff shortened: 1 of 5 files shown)


uw-frame-components/portal/notifications/partials/notifications-list-item.html
```diff
@@ -0,0 +1,23 @@
+<!-- NOTIFICATION CONTENT -->
+<a class="md-list-item-text"
+   ng-href="{{ notification.actionURL }}"
+   target="_blank"
+   aria-label="{{ notification.actionAlt }}"
+   layout="row"
+   layout-align="start center"
+   layout-xs="column"
+   layout-align-xs="center start">
+  <span><i class="fa fa-exclamation-triangle" ng-if="notification.priority"></i> {{ notification.title }}</span>
```
apetro: Priority-ness of notification is communicated only visually. `aria` considerations?

thevoiceofzeke: Priority notifications are also shown first in the list. I'm not sure the fact of its priority-ness adds anything helpful for screen reader users. :/

vertein: Maybe if we use ALL CAPS for priority notifications, screen readers will yell the content. ;)

PR #321, opened 2016-09-28
MUMMNG-2781 Update footer extension link

Doug-Reed:
Change footer link for UW-Extension help

### DIFF
--- a/uw-frame-components/js/frame-config.js
+++ b/uw-frame-components/js/frame-config.js
@@ -836,7 +836,7 @@ define(['angular'], function(angular) {
             "ariaLabelTitle" : "My U W",
             "crestalt" : "UW Extension",
             "group" : "UW System-Extension",
-            "footerLinks":[{ "url" : "https://kb.wisconsin.edu/myuwsystem/",
+            "footerLinks":[{ "url" : "http://ce.uwex.edu/technical-support/",
                 "target" : "_blank",
                 "title" : "Help"
               }],


timlevett:
Looking at the duplicate issue MUMMNG-2790, Will says hold off on doing this.

apetro:
To go down this road, I think we'd have to fork the `uw-extension` theme to have one theme for CEOEL and another theme for regular UW-Extension, and then the CEOEL fork would have the CEOEL-specific links (and maybe other branding too). They could still be at the same `my.uwex.edu` hostname as long as we were fancy about the group memberships driving which theme is selected.

I'm not opposed if you want to go this direction; it probably improves the user experience a little bit for CEOEL users. However, I think where we got in grooming is that

* CEOEL-specific content has inline CEOEL-specific help, that users are likely to find before they find the footer link.
* Will Crickman has worked to document the CEOEL support hand-off process for the Help Desk, so if CEOEL my.uwex.edu users do make contact, it'll  likely be handled appropriately.
* The number of CEOEL student users we're talking about currently is too small to justify this complexity. 
* UW-Extension et al. may eventually get to the product management insight that `my.uwex.edu` and `my.flex.uwex.edu` (or whatever the brand would be) are two distinct products that should have their own hostnames, their own brands, really be quite separated, and in that separation the footer links, themes, and everything about this will get simpler.

so I think the preferable way to go here is to ignore the CEOEL-specificity of support links for the time being and see how it shakes out.

Doug-Reed:
If the number of CEOEL users are too small to justify splitting Extension in two, I'll just go ahead and close this merge request and delete the branch.

________________________________

PR #323, opened 2016-10-04
Refactor notifications

thevoiceofzeke:
**In this PR**:
- Moved most of the logic handling notifications and their storage into the notifications service
- Added a service method that sorts notifications into dismissed/notDismissed arrays to be consumed by the controller, so the controller only has to worry about its two scope arrays
- Got rid of the "No [new/dismissed] notifications" message. Now, when there are no notifications of a type, that type's tab will simply disappear (similar to how the [outages site](http://outages.doit.wisc.edu/date/2016-09-23) doesn't show the "Unplanned" tab if there are no unplanned outages to show)
- Renamed some functions and variables to hopefully make their purposes more apparent to virgin eyes
- Added so many [jsdoc](http://usejsdoc.org/)-style comments to the service and controller that any person who can read English should be able to understand them
- Cleaned up some formatting/syntax inconsistencies, fixed some typos, added some semi-colons
- Removed jQuery from notifications controller (involved touching about 8 different files)
    - `has-priority-nots` class, which is responsible for styling of priority notifications, is now only applied to the `<portal-header>` directive -- this should help the related LESS make more sense

**Note:** I've been staring at this so long I have lost objectivity. ~~I cut out a lot of bloat but I'm only *pretty* sure none of it was needed, so please review carefully.~~ Upon further testing, I'm now quite certain none of it was needed. :)

### Demo of new functionality
notifications-demo

### Screenshots (desktop & mobile priority notifications)
screen shot 2016-10-04 at 2 38 09 pm
screen shot 2016-10-04 at 2 38 37 pm

### DIFF
--- a/uw-frame-components/css/buckyless/header.less
+++ b/uw-frame-components/css/buckyless/header.less
@@ -1,4 +1,5 @@
 portal-header {
+  display: block;
   a {
     &:not(.md-button):not(.btn):not(.launch-app-button) {
       &:hover, &:focus, &:active {
@@ -59,9 +60,12 @@ portal-header {
         }
       }
     }
+  }
+  &.has-priority-nots {
     @media (min-width: 768px) {
-      &.has-priority-nots {
-        margin-top: 46px;
+      margin-top: 46px;
+      md-toolbar {
+        top: 46px;
       }
     }
   }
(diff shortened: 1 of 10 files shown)


timlevett:
Might need to adjust the test case for the notification controller/service.

PR #324, opened 2016-10-04
Get third-party dependencies from CDN

paulerickson:
All of the third-party libraries used by uw-frame are available from CDNJS, so here's a suggestion to replace our use of bower dependencies with CDN URLs.

Note: adblockers like to block angulartics-google-analytics, so I made it fail gracefully in initialization rather than bork up the requirejs module loader or the angularjs "module" loader.

**Pros:**
* Browsers can use the same cached resource across all MyUW apps (or any site)
* Reduces build time
* Reduces build configuration
* Reduces load on webservers
* cdnjs has better uptime than we do
* All the URLs are versioned

**Cons:**
* Not managed through a command-line tool _(jspm+systemjs could address this)_
* A single page load might be slower, due to the extra DNS lookup
* Local webservers might be less latent for local users
* We don't control the CDN (I'm not sure this is a con at all, but it's a popular argument)


uw-frame-components/css/angular.less
```diff
@@ -1,6 +1,6 @@
-@import "../bower_components/normalize-less/normalize.less";
-@import "../../bower_components/font-awesome/css/font-awesome.css";
-@import "../bower_components/bootstrap/less/bootstrap.less";
+@import "https://raw.githubusercontent.com/additiveinverse/normalize.less/master/normalize.less";
```
timlevett: master?

uw-frame-components/portal/main.js
```diff
@@ -37,10 +37,15 @@ define([
...
-    'angulartics',
     'angulartics-google-analytics'
 ], function(angular, require) {
 
+    // Define a stub in case this angular module is undefined, i.e. was blocked
+    try {
+        angular.module('angulartics.google.analytics', []);
+    }
+    catch(e) {}
+
```
timlevett: clever.

timlevett:
I can edit my final review comment, this makes me sigh. I was going to add I really like this change so far, just a minor tweak needed for uw-frame-java

uw-frame-components/css/angular.less
paulerickson: Good catch - I tried to keep the same versions defined in bower, and in this case the version was '*'.  Should I just pick the latest tagged version?

package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "uw-frame",
-  "version": "2.8.0",
+  "version": "2.9.0",
```
vertein: We don't have to update the package.json, it's gets updated as part of our release process with a version that matches the java release as well.  Eventually, it would be cool to doing auto releasing.

uw-frame-components/css/angular.less
paulerickson: K, I confirmed that bower was pulling from master, and probably intentionally, since the last tag (the only tag), v0.1, contains a much older version, 2.1.3.  Basically, this looks like an abandoned fork with weird versioning.
I fixed the URL to the current (two year old) revision so that the build is deterministic.

paulerickson: I would not want to change dependencies in this same revision, but Bootstrap includes a normalize.less - is that just as good?

paulerickson:
Okay, I believe I have sorted out the pom and addressed all the feedback.  `npm run jetty` and `npm run static` both look good.

paulerickson:
False alarm, there's still more.

timlevett:
also may want to check `npm run docs`

paulerickson:
@timlevett Done.  I did the same with docs and made sure it loads.

Two notification service unit tests are failing and I can't see why so I ran git bisect.  It says they started failing in the last revision:
```
first bad commit: [d0aae5e15caa8ebde07f11d24fb572a1d2a809ec] Refactor notifications controller and service
```

PR #329, opened 2016-10-06
Clean up promises on notification service

timlevett:
+ Fixed bug where if you had dismissed notification ids in your array, the sortNotifications wouldn't know about them until it was to late (promises)
+ Added test case to verify the dismissed is populated properly.


timlevett:
I have to rethink this a little bit, looks like the getNotificationsByGroups() actually calls getAllNotifications() which already calls getDismissedNOtificationIds(). Will resolve and post a new commit. Maybe even with a test case :dancer:

PR #332, opened 2016-10-07
MUMUP-2716 : Model Mascot announcement end date in friendly format

Daha5:
Changing the services.js file to take in a iso time and then convert it back to epoch so it doesn't mess up the flow of the rest of the code.

Changes will be made to the features.json file accordingly to send in an iso time instead of epoch.

### DIFF
--- a/uw-frame-components/portal/features/services.js
+++ b/uw-frame-components/portal/features/services.js
@@ -238,7 +238,7 @@ define(['angular'], function(angular) {
               var startDate = Date.parse(new Date(feature.goLiveYear, feature.goLiveMonth, feature.goLiveDay));
               var expirationDate = feature.buckyAnnouncement.endDate;
               if (typeof expirationDate=="string"){
-                  var expirationDate = new Date(expirationDate).getTime();
+                  expirationDate = new Date(expirationDate).getTime();
               }
               
               if(startDate <= today && today <= expirationDate) {
--- a/uw-frame-components/portal/features/services.js
+++ b/uw-frame-components/portal/features/services.js
@@ -236,8 +236,10 @@ define(['angular'], function(angular) {
               //check dates
               var today = Date.parse(new Date());
               var startDate = Date.parse(new Date(feature.goLiveYear, feature.goLiveMonth, feature.goLiveDay));
-              var isoDate = feature.buckyAnnouncement.endDate;
-              var expirationDate = new Date(isoDate).getTime();
+              var epochDate = feature.buckyAnnouncement.endDate;
+              if (typeof epochDate=="string"){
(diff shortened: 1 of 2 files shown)


timlevett:
Suggest also changing https://github.com/UW-Madison-DoIT/uw-frame/blob/master/uw-frame-components/staticFeeds/features.json as part of this.

uw-frame-components/portal/features/services.js
```diff
@@ -236,7 +236,10 @@ define(['angular'], function(angular) {
               //check dates
               var today = Date.parse(new Date());
               var startDate = Date.parse(new Date(feature.goLiveYear, feature.goLiveMonth, feature.goLiveDay));
-              var expirationDate = feature.buckyAnnouncement.endDate;
+              var isoDate = feature.buckyAnnouncement.endDate;
+              var temp = new Date(isoDate);
+              var expirationDate = temp.getTime();
```
apetro: This is fine. I'd be tempted to avoid the `temp` and do `var expirationDate = new Date(isoDate).getTime();`.

uw-frame-components/staticFeeds/features.json
```diff
@@ -37,7 +37,7 @@
     "buckyAnnouncement" : {
       "shortTitle" : "Try the new MyUW search",
       "shoreDesc" : "Now search across MyUW, Wisc.edu, and the campus directory.",
-      "endDate": 1487952401000
+      "endDate": "2017-02-24T10:06:41-0600"
```
vertein: We can take off the T:.....

uw-frame-components/portal/features/services.js
```diff
@@ -236,7 +236,9 @@ define(['angular'], function(angular) {
               //check dates
               var today = Date.parse(new Date());
               var startDate = Date.parse(new Date(feature.goLiveYear, feature.goLiveMonth, feature.goLiveDay));
-              var expirationDate = feature.buckyAnnouncement.endDate;
+              var isoDate = feature.buckyAnnouncement.endDate;
```
apetro: How messy would it be to support both epoch time and ISO date in `feature.buckyAnnouncement.endDate`? As in, maybe, if it's an integer, it's an epoch time, but if it's an ISO date, well, then it's an ISO date?

Asking for a couple reasons.

1. Would ease the upgrade path, not having to update `features.json` in the tiers in the same transaction with upgrading to a `uw-frame` version with this changeset.
2. Would ease the upgrade path for non-MyUW adopters of `angularJS-portal`.

uw-frame-components/portal/features/services.js
thevoiceofzeke: ^Seconded

PR #335, opened 2016-10-11
Simplify build

jsamsa:
This change was to move functionality found in build.sh files and Grunt configuration into the npm build with the goal of simplifying the build process by reducing the number of tools required.


vertein:
I like it.  Does anything change in our documentation here: https://github.com/UW-Madison-DoIT/uw-frame/blob/master/README.md or  https://github.com/UW-Madison-DoIT/uw-frame/blob/master/docs/markdown/quickstart.md which results in the io page here being updated: http://uw-madison-doit.github.io/uw-frame/latest/#/md/quickstart
?

jsamsa:
- Fixed missing rimraf issue
- Updated documentation
- Modified Dockerfile

Dockerfile
```diff
@@ -1,17 +1,19 @@
 FROM node:5.7.1-slim
 MAINTAINER Tim Levett <tim.levett@wisc.edu>
-RUN npm install -g grunt-cli superstatic
+RUN npm install -g superstatic
 
 RUN apt-get update && apt-get install -y bzip2 && apt-get install -y git
 
 # Add frame
+COPY tools /build/tools
+copy docs /build/docs
```
paulerickson: uppercase `COPY` is more [conventional](https://docs.docker.com/engine/reference/builder/#/format)

paulerickson:
:+1: less is more!

PR #336, opened 2016-10-12
MUMUP-2761 : add ability to have a name for the default theme

### LINKED ISSUE #333: Fails to load when a theme is specified with uw-frame 2.9.0
Working on uw-digital-id-app, I had OVERRIDE.APP_FLAGS.defaultTheme set to 'uw-system', but when I upgraded from uw-frame 2.8.0 to 2.9.0 it stopped loading.

Javascript throws an `Uncaught TypeError: Cannot read property 'name' of undefined` from uw-frame-components/portal/main.js:176
```javascript
174       var generateTheme = function(name) {
175         if(!name) {
176           name = $rootScope.portal.theme.name;
177         }
178         $rootScope.portal.theme.themeVersion = THEMES.themeVersion;
179         var mdTheme = $mdTheming.THEMES[name];
180         if(mdTheme) {
181           $mdTheming.generateTheme(name,null);
182         }
183       };
```

screenshot from 2016-10-10 10-58-29

As a workaround, I'm just removing the defaultTheme configuration from my app.

timlevett:
Resolves #333

You can now have a name as `APP_FLAG.defaultTheme`. Before you could just have an index number or the string `'group'`.

### DIFF
--- a/docs/markdown/configuration.md
+++ b/docs/markdown/configuration.md
@@ -3,7 +3,7 @@ Your application can overwrite any constant listed below by adding it to the `js
 ### Configuration options
 
 #### APP_FLAGS
-+ `defaultTheme` : This is the default theme you want (see frame-config.js for the array list of themes). Provide an index number to just have simple selection, or set to the string `'group'` to enable group selection. If you do group selection make sure you set the `SERVICE_LOC.groupURL`.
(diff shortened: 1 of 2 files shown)


paulerickson:
:+1:

PR #337, opened 2016-10-13
Break the dependency on docs submodule

jsamsa:
The Gruntfile had an implied dependency coupling static and doc project
in the processhtml target. In practice, it appears Grunt would disregard
the missing docs directory during the docker build, but the npm
htmlprocessor call which I added would not, so I added the docs module
in the docker build. This reverts that dependency.

See Gruntfile.js in commit ab764d6d1695069a288f96a70e8dd84fe5171c23


jsamsa:
Put more simply, the Docker container didn't have a dependency on "docs", I added it to fix the build, this change removes that unnecessary dependency.

vertein:
:+1:

PR #339, opened 2016-10-20
Upgrade uw-frame to Angular Material 1.1.1

thevoiceofzeke:
**In this PR**:
- Updated to latest version of Angular Material
- Various CSS changes to compensate for said update
- Pruned LESS files for unused styles 
- Moved all styles related to exclusive mode portlets into their own file
- Replaced hard-coded media queries with LESS variables
- Redesigned mobile search functionality due to breaking change in angular material

### Search demo
search-demo

*Note: Andrew pointed out that the search field should auto-focus when it's visible. I agree. There's a bit more technical work to do to get that working, so I'd like to move forward with this at the moment.*


uw-frame-components/css/buckyless/helpers.less
```diff
@@ -1,58 +0,0 @@
-/* General Helper Classes */
```
apetro: How do I think about things like this disappearing in the spirit of [Semantic Versioning](http://semver.org/)? Losing styles can be not-backwards-compatible as experienced by frame-based apps, such that they'd have to change stuff upon upgrade to a version of frame with this changeset? Does that mean this changeset should bump the `MAJOR` version of frame?

apetro:
On `predev`, when searching and awaiting result loading, the loading indicator displays too far left.

loading spinner is too far left

apetro:
Search on mobile dismisses itself when I try to tap into the search query box.

you-cant-search-this

apetro:
Weirdness with mascot hover tip when page is scrolled down or scrolls.

scrolling_tip

thevoiceofzeke:
The tooltip thing is actually part of angular material. You can see the same issue if you scroll quickly on their [demo page](https://material.angularjs.org/1.0.9/demo/tooltip). It's more pronounced in bucky's case because the top bar is fixed, so he retains his hover state. There might be something we can do about this, but it's not something that was caused by the upgrade to 1.1.1.

thevoiceofzeke:
Pushed a fix for the loading gif issue in [angularjs-portal #551](https://github.com/UW-Madison-DoIT/angularjs-portal/pull/551)

thevoiceofzeke:
@apetro: Regarding the theme selector, see [Angular Material's select demo](https://material.angularjs.org/latest/demo/select) (the "State" drop down). It seems that style is intentional (and only present when the select is in a empty state).

uw-frame-components/portal/settings/partials/settings.html
```diff
@@ -10,16 +10,19 @@
...
       </div>
-      <!-- Theme selector and 'danger zone' -->
+      <!-- Theme selector and "danger zone" -->
       <div layout="column" flex-gt-sm="40">
-        <h4 class="md-subhead">Theme Selector<br><small>Pick which place you call home.</small></h4>
-        <select class="form-control" ng-options="item as item.name for item in THEMES track by item.name" ng-model="portal.theme">
-        </select>
+        <h4 class="md-subhead">Theme Selector</h4>
+        <md-input-container>
+          <md-select ng-model="portal.theme" placeholder="Pick which place you call home" aria-label="Select a theme">
```
apetro: I agree this behaves the same as in the [upstream demo](https://material.angularjs.org/latest/demo/select) and I agree it's not a dealbreaker for moving forward. 

I still say it feels bugged.  Clicking through many angular-material open issues about `select` I didn't see precisely this issue, though I saw a couple conceptually close to it. Anyway, I'm fine with our strategy of adopting widely-adopted technology so that if this is a real issue, well, it'll annoy lots of adopters and someone will fix it. :smile:

PR #340, opened 2016-10-20
Local link less

jsamsa:
This took more reorganizing than I hoped. My main concern was removing the ${project.version} component to the less compilation output files. I'm not sure if that was actually useful in any way and needed to go for me to consolidate less compilation into one place (the npm build). If Maven _must_ compile less for some reason that could be accomplished using the exec plugin.


paulerickson:
I'm also concerned about removing the version, since that would mean users don't pick up updates consistently.  It highlights an important gap between the static project and the java project, though.

jsamsa:
Cache busting not implemented but these changes should get past the less compilation failures. If anyone can point me to how I can pull the docker image travis-ci uses to run the tests, that would be useful. It seems to be using Ubuntu Precise, but it's not clear where to source it from.

vertein:
Hey!  I've been busy, but hoping to get some time to work with this today.  Docker build is using precise.  I don't know exactly where it's pulling from, I need to take a look at the configs.

We do need the project version.  Added in for caching fixes that were causing a lot of issues.  Also, maven does need to compile the less because of our build processes.

jsamsa:
@vertein would maven executing an npm target suffice? Also, can I assume we need cache busting but not necessarily the version number? I don't want to rule out options to solve the problem.

vertein:
Definitely agree about version number not being the only solution.

I need to take a look at our servers and see if npm is installed, if not I'm sure it could be.  Seems like a fine way to do it.

jsamsa:
@vertain https://github.com/eirslett/frontend-maven-plugin might be a preferred approach over having npm on the servers. This plugin provides isolation from a system-wide npm. 

Another, possibly more palatable option would be to publish the component module as a zip archive to a maven repository, then have the uw-frame-java build depend on that archive.

vertein:
That front end plugin looks cool.  Is there a real strong reason why we wouldn't continue to let Maven compile the less still and keep it as it is?  Is it just that we want to consolidate the build processes?

jsamsa:
It was mostly practical, which I'll explain, but in principle eliminating the redundancy and allowing the web components to be built independently seemed like good guidance. In practice, the npm modules for font-awesome, bootstrap, and normalize have a slightly different layout than the bower versions of these modules. The static build compiled less without a problem, but the maven plugin was having a difficult time referencing relative filenames. I think getting maven to compile with these can be done, but was faster to eliminate the one that seemed, to me, to be an outlier. 

Perhaps my biggest difficulty is walking into these changes without understanding all the complexity behind additional build use cases. Longer term, I could see 3 (or more) separate projects that are no longer nested: web-components, uw-frame-java (depends on web-components), uw-frame-static (depends on web-components). That would effectively decouple the build into 3 simpler builds and make it easier to implement additional types of containers in the future (java, static, ???). This change was a step in that direction.

paulerickson:
Travis is already running npm commands, so it's definitely available in the build environment.

tools/uw-frame-components/build.js
```diff
@@ -0,0 +1,28 @@
...
+var less = require('less');
+var exec = require('child_process').exec;
+var glob = require('glob');
+
+var exec_handler = function(error, stdout, stderr) {
+    if (error) throw error;
+}
+
+exec('mkdir -p target/uw-frame-components', exec_handler);
+exec('cp -r uw-frame-components/* target/uw-frame-components', exec_handler);
```
jsamsa: @paulerickson agreed. It will take a little effort to ensure the java and static versions can work without much if any variation.

Issue #338, opened 2016-10-19
Static build is critically broken (development blocker)


thevoiceofzeke:
As of #335, running uw-frame static locally is broken to the point that it is a major blocker for development. 
### Build fails

When running `npm run static:dev`, the build will fail ~60% of the time and log the following error:

``` bash
Error: Command failed: /bin/sh -c node_modules/.bin/lessc -x uw-frame-components/css/themes/uw-river-falls.less > uw-frame-static/target/css/themes/uw-river-falls.css
FileError: resource 'https://cdn.rawgit.com/twbs/bootstrap/v3.3.6/less/mixins/opacity.less' gave this Error:
  Error: getaddrinfo ENOTFOUND cdn.rawgit.com cdn.rawgit.com:443
 in https://cdn.rawgit.com/twbs/bootstrap/v3.3.6/less/mixins.less on line 6, column 1:
5 @import "mixins/hide-text.less";
6 @import "mixins/opacity.less";
7 @import "mixins/image.less";


    at ChildProcess.exithandler (child_process.js:213:12)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:172:7)
    at maybeClose (internal/child_process.js:821:16)
    at Socket.<anonymous> (internal/child_process.js:319:11)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at Pipe._onclose (net.js:469:12)

npm ERR! Darwin 16.0.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "run" "static:dev"
npm ERR! node v5.4.0
npm ERR! npm  v3.3.12
npm ERR! code ELIFECYCLE
npm ERR! uw-frame@2.9.0 static:dev: `npm run build-static -s && concurrently -k "npm run static -s" "npm run static:watch -s"`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the uw-frame@2.9.0 static:dev script 'npm run build-static -s && concurrently -k "npm run static -s" "npm run static:watch -s"'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the uw-frame package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     npm run build-static -s && concurrently -k "npm run static -s" "npm run static:watch -s"
npm ERR! You can get their info via:
[...]

PR #342, opened 2016-10-26
Issue 338 - Fixes cdn error and uses npm for less compiling

vertein:
I also ran into https://github.com/UW-Madison-DoIT/uw-frame/issues/338 about 70% of the time when I was on wireless, but almost 0% of the time when I was wired.

This also takes the idea of compiling all the less in one spot from : https://github.com/UW-Madison-DoIT/uw-frame/pull/340/ while keeping the version numbering that we are currently using.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,9 +2,6 @@
   "name": "uw-frame",
   "version": "2.9.0",
   "description": "MyUW's frame project for creating angular based microservices.",
-  "config" : {
-    "javaversion" : "1"
-  },
   "scripts": {
     "test": "karma start uw-frame-components/karma.conf.js --single-run",
     "posttest": "cd uw-frame-components",
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "uw-frame",
   "version": "2.9.0",
   "description": "MyUW's frame project for creating angular based microservices.",
+  "config" : {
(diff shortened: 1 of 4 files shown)


vertein:
Failing travis ci builds also gave me pause.  There were two issues.  1. fixed maven pom to point at directory where package.json was.  2. Fixed a cp error in build.js where if directory already existed, the build would failed.  A couple of versions ago we were removing that directory explicitly before trying to copy over.  This goes back to that.

package.json
```diff
@@ -2,13 +2,17 @@
   "name": "uw-frame",
   "version": "2.9.0",
   "description": "MyUW's frame project for creating angular based microservices.",
+  "config" : {
+    "javaversion" : "1"
```
apetro: What's `javaversion`? Under what circumstances ought it to be incremented? How would I know that?

paulerickson:
npm runs maven runs npm... I think that a circular dependency like this makes the project more difficult to understand and maintain

vertein:
@paulerickson I do too.  I was/am actually fine with maven compiling on their own less (https://github.com/UW-Madison-DoIT/uw-frame/pull/340#issuecomment-255415981).

I'd like to look into removing a lot of the npm calling maven things, like the clean scripts and the build scripts in the future.

PR #344, opened 2016-11-10
MUMUP-2705 Address href noopener and no referrer vulnerability

vertein:
PR to go hand in hand with https://github.com/UW-Madison-DoIT/angularjs-portal/pull/552


uw-frame-components/portal/misc/partials/circle-button.html
```diff
@@ -1,5 +1,6 @@
 <md-button ng-href="{{ cbDisabled ? '#' : href}}"
            target="{{cbDisabled ? '' : target}}"
+           rel='noopener noreferrer'
```
apetro: Are there negative consequences of *always* using `noopener noreferrer` even when `target` is *not* `_blank`?

PR #348, opened 2016-11-22
Removes the rm that caused inital build failures

vertein:
On an initial clone of the repo, the rm would fail (even though it has a -r).  Removes this line as someone could run `npm run clean` if they wanted to do a clean build.

### DIFF
--- a/tools/uw-frame-static/build.js
+++ b/tools/uw-frame-static/build.js
@@ -10,8 +10,6 @@ var exec_handler = function(error, stdout, stderr) {
   if (error) throw error;
 };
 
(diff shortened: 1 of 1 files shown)


tools/uw-frame-static/build.js
```diff
@@ -10,8 +10,6 @@ var exec_handler = function(error, stdout, stderr) {
   if (error) throw error;
 };
 
-exec('rm -rf uw-frame-static/target/', exec_handler);
```
apetro: In lieu of this, should it go ahead and run `clean`?

vertein: I like the separation of tasks.

PR #350, opened 2016-11-22
Upgrade angular-ui-bootstrap and angular libraries

jsamsa:
The angular-ui-bootstrap upgrade moves up 2 major versions and does
break compatibility with some older components. In particular
typeahead needed to switch the typeahead attribute to uib-typeahead.


jsamsa:
Note, it may be worth considering a major version bump for this since it does break compatibility with some ui components.

PR #360, opened 2017-01-13
MUMUP-2798 UW-Frame build works for windows

vertein:
Allows uw-frame to built on windows.

Removes the unix and maven specific commands from npm.  No more npm calling maven.


uw-frame-java/pom.xml
```diff
@@ -218,43 +180,277 @@
...
+        <version>1.0.1</version>
+        <executions>
+          <execution>
+            <id>rename-uw-madison-file</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>rename</goal>
+            </goals>
+            <configuration>
+              <sourceFile>${basedir}/target/frame/css/themes/uw-madison.css</sourceFile>
```
apetro: This cries out for refactoring to avoid hard-coding the set of themes in this `pom.xml`. Some kind of file name pattern matching solution?

vertein: I admittedly tried for a while and wasn't able to do it.  MUMUP-2810 to track this.  This could also be a step in the right direction to allow us to be a little bit less UW specific.

PR #361, opened 2017-01-25
MUMUP-2828 Adopt Jekyll/GitHub Pages for documentation

apetro:
Naive conversion of docs to build using Jekyll as implemented in GitHub pages.

Has the effect of "fixing" the `uw-frame` documentation, resolving the hanging forever loading bug.

uw-frame docs

([Live preview](https://apetro.github.io/uw-frame/)).

Renders, so it's better than broken docs.

Much opportunity for further improvement.

The live loading indicator etc. examples are orphaned by this change -- probably should move that into `tools` or even directly into `AngularJS-portal` as available on `public.my`.


apetro:
I already adjusted the GitHub Pages settings on this repo to build from the `master` branch `docs` folder rather than the `gh_pages` branch, so when this PR merges the docs site should update to reflect.

PR #367, opened 2017-02-02
Don't repeat README badges in docs website, except Gitter.

apetro:
Include Gitter in context with other ways to get in touch.


apetro:
Nevermind - let's see how we like the new site for a while and tweak from there.

PR #385, opened 2017-03-29
Allow widgets to be embedded in frame apps

thevoiceofzeke:
[MUMUP-2874](https://jira.doit.wisc.edu/jira/browse/MUMUP-2874)

**In this PR:**
- Widget styles and directives ported from angularjs-portal
- Frame apps can now embed existing widgets by using the new widget directive
    e.g. `<widget fname="weather"></widget>`
- Added a service to fetch the JSON for a full entity file (given `fname` attribute), get additional JSON from URLs provided in the entity file, and get RSS feeds as JSON
- Attempted to give widget directives and files more sensible, descriptive names and comments than their angularjs-portal counterparts
- Added more error logging (via angular's `$log`) and replaced some instances of `console.log`
- Simplified some service calls and widget controller functions
- For demoing purposes: Created demo page in uw-frame static and a few sample widget JSON files
- Syntax fixes, code formatting, etc.

*Note: Because we opted to do this the right way, rather than attempting a workaround that would be compatible with angularjs-portal, the predev build of angularjs-portal is going to be very buggy until that app can be refactored to completely decouple its widget and layout functionality and instead rely upon the widget functionality now provided by uw-frame.*

### Screenshot (static demo example)
screen shot 2017-03-29 at 1 31 02 pm


uw-frame-components/my-app/main.js
```diff
@@ -21,8 +29,9 @@ define(['angular', 'jquery', 'portal', 'portal/main/routes', 'portal/settings/ro
             when('/access-denied', main.accessDenied).
             when('/server-error', main.serverError).
             when('/sorry-safari', main.storageError).
+            when('/demo-widgets', widgets.demoWidgets).
             when('/', main.main).
-            otherwise(main.fileNotFound); //default to 404
+            otherwise('/');
```
vertein: :+1:

uw-frame-components/portal/main.js
```diff
@@ -331,7 +338,7 @@ define([
         });
       }
 
-      //List of config. VERY IMPORTANT THAT THE CONFIGS has a corrisponding configsName in the same index
+      // List of config. VERY IMPORTANT THAT THE CONFIGS has a corresponding configsName in the same index
```
vertein: :+1:

uw-frame-java/src/main/webapp/WEB-INF/web.xml
```diff
@@ -26,5 +26,7 @@
       <url-pattern>/about</url-pattern>
       <url-pattern>/access-denied</url-pattern>
       <url-pattern>/server-error</url-pattern>
+      <url-pattern>/sorry-safari</url-pattern>
```
vertein: :+1:

PR #387, opened 2017-03-30
MUMUP-2875 Fixes bug with dataObject in notification feed caused it to not display

vertein:
Fixes a bug where the data object to find to show notification would return but we were checking if length > 0.  Well, length>0 is only valid when it's an array, so if it was just an object, it would fail and not show the notification.  Added test case as well.

----

### DIFF
--- a/uw-frame-components/portal/notifications/services.js
+++ b/uw-frame-components/portal/notifications/services.js
@@ -129,15 +129,15 @@ define(['angular', 'jquery'], function(angular, $) {
                   return;
                 }
                 var arrayFilter = angular.fromJson(notification.dataArrayFilter);
-                objectToFind = $filter('filter')(objectToFind, arrayFilter);
-              }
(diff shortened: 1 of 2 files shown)


uw-frame-components/portal/notifications/services.js
```diff
@@ -129,15 +129,15 @@ define(['angular', 'jquery'], function(angular, $) {
                   return;
                 }
                 var arrayFilter = angular.fromJson(notification.dataArrayFilter);
-                objectToFind = $filter('filter')(objectToFind, arrayFilter);
-              }
+                if ($filter('filter')(objectToFind, arrayFilter).length > 0) {
+                  return notification;
+                }
+              } else if (objectToFind) {
+                return notification
```
ChristianMurphy: :art: could this statement be closed with a semicolon?

vertein: I don't know how I could have missed that!

PR #393, opened 2017-04-12
MUMUP-2898 Adds ability to write your own directive for removing a widget

vertein:
As far as the app-framework is concerned, widgets don't need to be removed, they only need to be rendered.  Removal is more for apps that have layouts (like uPortal-home), so this allows the apps using the app-framework to have a removal button and decide what the removal actually does for widgets.

----


uw-frame-components/portal/widgets/partials/type__option-link.html
```diff
@@ -3,7 +3,7 @@
   <!-- OPTION-LINK ICON -->
   <div class="option-link-icon">
     <a href="{{ widget.selectedUrl }}" target="_blank" rel="noopener noreferrer">
-      <frame-widget-icon></frame-widget-icon>
+      <widget-icon></widget-icon>
```
thevoiceofzeke: 🤦‍♂️

PR #395, opened 2017-04-13
Add sorter function to allow sorting strings as numbers

thevoiceofzeke:
**In this PR:**
- Supports ordering leave balances by value

### DIFF
--- a/uw-frame-components/portal/widgets/controllers.js
+++ b/uw-frame-components/portal/widgets/controllers.js
@@ -260,7 +260,7 @@ define(['angular'], function(angular) {
     };
 
     /**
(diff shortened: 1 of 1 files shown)


uw-frame-components/portal/widgets/controllers.js
```diff
@@ -260,33 +260,37 @@ define(['angular'], function(angular) {
...
         return [];
       }
     };
 
     /**
+     * Used in tandem with filteredArray to display Leave Balances in order of value
+     * @param item Object containing a name string and a value string
+     * @returns {Number}
+     */
+    $scope.sortStringAsNumber = function(item) {
```
davidmsibley: where is this being used?

PR #396, opened 2017-04-13
Remove constraints to widget width and leave it up to containing element

thevoiceofzeke:
Fixes this:

screen shot 2017-04-13 at 1 14 23 pm

Decisions about how wide widgets can/should be should be left up to the people implementing widgets. It's up to individual projects to define containing elements and layouts.


Doug-Reed:
:+1:

PR #402, opened 2017-04-18
Launch behavior should default to same tab

thevoiceofzeke:
[MUMMNG-3294](https://jira.doit.wisc.edu/jira/browse/MUMMNG-3294)

**In this PR:**
- Changes the default behavior for widget launch buttons to open in same window/tab


apetro:
Consider enhancing `directives.md` to document the default behavior when the `target` attributes aren't set on the directives.

PR #403, opened 2017-04-19
Fix RSS widget display wonkiness

thevoiceofzeke:
[MUMUP-2913](https://jira.doit.wisc.edu/jira/browse/MUMUP-2913)

**In this PR**:
- Add styles to stop RSS titles from wrapping
- Don't show dates if they're undefined (currently, the [rssToJson](https://github.com/UW-Madison-DoIT/rssToJson/blob/master/src/main/java/edu/wisc/my/rssToJson/service/RsstoJsonServiceImpl.java) doesn't set any pubDate, so it's always coming back undefined. Until it's fixed, titles should be displayed at full width 
- Do not underline titles on hover
- Adjust default limit of title length depending on whether dates are displayed
- Leverage launch-button directive within RSS widgets
- Trim leading/trailing spaces from RSS titles before truncating them
- Adjust appearance of widgets demo page in static

screen shot 2017-04-19 at 11 41 44 am

### DIFF
--- a/uw-frame-components/portal/widgets/controllers.js
+++ b/uw-frame-components/portal/widgets/controllers.js
@@ -149,8 +149,12 @@ define(['angular'], function(angular) {
      */
     $scope.getPrettyDate = function(dateString) {
       // Create a new date if a date string was provided, otherwise return null
-      $log.log('rss widget got date string: ' + dateString);
-      return dateString ? new Date(dateString) : null;
+      if (dateString === undefined) {
+        $log.warn('Date string in RSS widget controller is undefined');
+        return null;
+      } else {
+        return new Date(dateString);
+      }
     };
 
     /**
@@ -173,11 +177,7 @@ define(['angular'], function(angular) {
         $scope.config.lim = 5;
       }
       if (!$scope.config.titleLim) {
-        if ($scope.config.showdate) {
-          $scope.config.titleLim = 35;
-        } else {
-          $scope.config.titleLim = 45;
-        }
+        $scope.config.titleLim = 45;
       }
       if (!$scope.config.showShowing) {
         $scope.config.showShowing = false;
--- a/uw-frame-components/portal/widgets/controllers.js
+++ b/uw-frame-components/portal/widgets/controllers.js
@@ -149,6 +149,7 @@ define(['angular'], function(angular) {
      */
     $scope.getPrettyDate = function(dateString) {
       // Create a new date if a date string was provided, otherwise return null
+      $log.log('rss widget got date string: ' + dateString);
       return dateString ? new Date(dateString) : null;
     };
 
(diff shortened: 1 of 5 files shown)


uw-frame-components/portal/widgets/partials/demo-widgets.html
```diff
@@ -5,33 +5,29 @@
...
+    <!--
+      LOCAL DEMO
+      - Must set SERVICE_LOC->widgetApi.entry in app-config.js to: "staticFeeds/"
+    -->
+    <div layout="row" layout-align="center center">
+      <div flex-xs="100" style="max-width:310px">
         <widget fname="sample-widget__search-with-links"></widget>
       </div>
-      <div flex-xs="100">
+      <div flex-xs="100" style="max-width:310px">
```
ChristianMurphy: Could this inline styles be moved into the stylesheet?

thevoiceofzeke: The only time it should ever be applied is on that specific page, which is only used when running locally. I'd rather not muddy the stylesheet with a class for this singular use case.

uw-frame-components/portal/widgets/controllers.js
```diff
@@ -149,7 +149,21 @@ define(['angular'], function(angular) {
      */
     $scope.getPrettyDate = function(dateString) {
       // Create a new date if a date string was provided, otherwise return null
-      return dateString ? new Date(dateString) : null;
+      if (dateString === undefined) {
```
davidmsibley: should this explicitly be checking undefined? or just falsy?

davidmsibley: I bring it up cause now that we've added the getPublishedDate stuff in the rssToJson, it's likely an item will come back with a pubDate of null

uw-frame-components/portal/widgets/partials/type__rss.html
```diff
@@ -5,10 +5,10 @@
   <ul class="widget-list">
     <li ng-repeat="item in data.items | filter:filterText | limitTo:config.lim">
       <a class="full-width" ng-href="{{ item.link }}" target="_blank" rel="noopener noreferrer">
-        <div class="headline" ng-class="{ 'nodate' :  !config.showdate }">
-          {{ item.title | truncate:config.titleLim }}
+        <div class="headline" ng-class="{ 'nodate' :  !config.showdate || item.pubDate === undefined }">
```
davidmsibley: same falsy as above

uw-frame-components/portal/widgets/controllers.js
thevoiceofzeke: Eh, yeah. It should probably just be reverted back to the old ternary. I made the change before I was planning on changing rssToJson :P

uw-frame-components/portal/widgets/partials/type__rss.html
thevoiceofzeke: The check can just be removed here, now that rssToJson has been updated.

PR #409, opened 2017-04-24
Format dates consistently across all RSS widgets

thevoiceofzeke:
Allowing widget creators the ability to specify a date format could create an inconsistent appearance. Sticking to a rigid format is the best way to ensure uniform appearance across all RSS widgets. 

Also in this PR:
- Don't let widget config set limits that are too long

### DIFF
--- a/uw-frame-components/portal/widgets/controllers.js
+++ b/uw-frame-components/portal/widgets/controllers.js
@@ -181,6 +181,13 @@ define(['angular'], function(angular) {
       if (!$scope.config.showShowing) {
         $scope.config.showShowing = false;
       }
+      // Set sensible fallbacks in case widget config would create display problems.
+      if ($scope.config.lim > 6) {
+        $scope.config.lim = 6;
+      }
+      if ($scope.config.titleLim > 50) {
+        $scope.config.titleLim = 50;
+      }
     };
 
     /**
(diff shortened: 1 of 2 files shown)


uw-frame-components/portal/widgets/controllers.js
```diff
@@ -181,6 +181,13 @@ define(['angular'], function(angular) {
       if (!$scope.config.showShowing) {
         $scope.config.showShowing = false;
       }
+      // Set sensible fallbacks in case widget config would create display problems.
+      if ($scope.config.lim > 6) {
+        $scope.config.lim = 6;
+      }
+      if ($scope.config.titleLim > 50) {
```
ChristianMurphy: :thought_balloon: Should this be checked in a JSON schema type validation, instead of at run time?

thevoiceofzeke:
@ChristianMurphy I updated to use angular's localizable date formatting. I'd really rather not include the year, but angular doesn't offer a localizable format without it :(

uw-frame-components/portal/widgets/controllers.js
thevoiceofzeke: This is just a stop-gap -- not worth spending more than a couple minutes on since it's not currently needed and not in scope of any story.

ChristianMurphy: Fair enough,
It would be good to have a discussion about long term configuration management.
Mixing configuration validation into business logic into is a easy stop gap to add, but can be difficult to track down an remove down the road.

PR #411, opened 2017-04-24
Use launch button directive and default launch text in all widget types

thevoiceofzeke:


uw-frame-components/portal/widgets/partials/type__list-of-links.html
```diff
@@ -27,8 +27,7 @@
...
 
 <!-- LAUNCH BUTTON -->
-<launch-button ng-if='config.launchText && widget.url'
-               data-href="{{widget.url}}"
+<launch-button data-href="{{widget.url}}"
                data-target="{{ widget.target ? widget.target : '_self' }}"
-               data-button-text="{{config.launchText}}"
-               data-aria-label="{{'launch ' + config.launchText}}"></launch-button>
+               data-button-text="{{ config.launchText ? config.launchText : 'Launch full app' }}"
+               data-aria-label="{{ config.launchText ? config.launchText : 'Launch ' + widget.title }}"></launch-button>
```
ChristianMurphy: I understand why this is being done, and that it may be the best that can be done right now.
This is another place where it would be nice to replace logic in the template with default values passed back from a configuration validator.
E.G. [json schema default](https://spacetelescope.github.io/understanding-json-schema/reference/generic.html#metadata)

PR #414, opened 2017-04-27
Add default username menu "avatar" with feature flag

thevoiceofzeke:
[MUMUP-2903](https://jira.doit.wisc.edu/jira/browse/MUMUP-2903)

>As a non-Madison user, I would like profile avatar to have sensible defaults, so that MyUW feels more like it was designed with me in mind than like I'm an afterthought in a Madison-centric product.

**In this PR:**
- Bare minimum stub of what the avatar feature might look like 
- Default behavior modeled after Google's existing convention for users without profile photos
- Explicit colors for each theme (necessary because we can't rely on material theming for every theme, nor can we do across-the-board coloring for all themes)
- Feature flag to enable avatar. For now, this only differentiates the existing functionality from the new default "avatar." In the future, this will be the toggle to show a real profile image vs. the default.

### Screenshots from various themes
screen shot 2017-04-27 at 11 40 46 am
screen shot 2017-04-27 at 11 41 00 am
screen shot 2017-04-27 at 11 41 09 am
screen shot 2017-04-27 at 11 41 17 am
screen shot 2017-04-27 at 11 41 26 am
screen shot 2017-04-27 at 11 41 49 am
screen shot 2017-04-27 at 11 42 17 am
screen shot 2017-04-27 at 11 53 01 am
screen shot 2017-04-27 at 11 53 08 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,15 +34,17 @@
   ],
   "contributors": [
     {
-      "name": "Tim Levett",
-      "email": "tim.levett@wisc.edu"
-    },
-    {
       "name": "Timothy Vertein"
     },
     {
-      "name": "Jared Hanstra",
-      "email": "jared.hanstra@wisc.edu"
+      "name": "Doug Reed"
+    },
+    {
+      "name": "David Sibley"
+    },
+    {
+      "name": "Zeke Witter",
+      "email": "david.witter@wisc.edu"
     }
   ],
   "bugs": {
--- a/uw-frame-components/portal/main/controllers.js
+++ b/uw-frame-components/portal/main/controllers.js
@@ -72,19 +72,26 @@ define(['angular','require'], function(angular, require) {
 
   /* Username */
   app.controller('SessionCheckController', [ '$scope', 'mainService', 'NAMES', 'FOOTER_URLS', '$rootScope', function($scope, mainService, NAMES, FOOTER_URLS, $rootScope) {
-    var that = this;
-    that.user = [];
+    var vm = this;
+    vm.user = [];
+    vm.username = '';
+    vm.firstLetter = '';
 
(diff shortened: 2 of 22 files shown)


package.json
```diff
@@ -34,15 +34,17 @@
   ],
   "contributors": [
     {
-      "name": "Tim Levett",
```
apetro: I wouldn't remove acknowledging @timlevett . He's an appreciated historical contributor. If there's an emeritus status flag or the like, I think that'd be appropriate, but otherwise I'd just leave him in the `contributors` array, figuring retaining him doesn't harm anything.

vertein: `contributors` is a good call.  He doesn't at this point have publish privs to npm for the project.

uw-frame-components/portal/main/controllers.js
```diff
@@ -72,19 +72,25 @@ define(['angular','require'], function(angular, require) {
...
-      if (NAMES.guestUserName && that.user && that.user.userName === NAMES.guestUserName)
+      vm.user = result;
+      // Check if is guest
+      if (NAMES.guestUserName && vm.user && vm.user.userName === NAMES.guestUserName) {
         $rootScope.GuestMode = true;
-
+      } else {
+        // Get first letter of first name or display name
+        var username = vm.user.firstName ? vm.user.firstName : vm.user.displayName;
+        vm.firstLetter = username.substring(0,1);
```
apetro: Could we make this fail safe to, say, "?" in the case where `username` ended up `nil` or `""`.

It'll probably never actually happen, but user attributes, they can get weird.

package.json
thevoiceofzeke: So what are you suggesting, Tim? Leave him in the array?

PR #418, opened 2017-04-28
Compact widget directive

vertein:
Moves more widget code from uPortal-home to the uPortal-App-Framework, so everyone can enjoy a good compact mode widget.

Fixes a bug in uPortal-home where compact mode wasn't rendering because of switching to the new layout call in portal

----

- [X] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

### DIFF
--- a/uw-frame-components/portal/widgets/partials/compact-widget-card.html
+++ b/uw-frame-components/portal/widgets/partials/compact-widget-card.html
@@ -6,13 +6,6 @@
     <md-icon>info</md-icon>
   </md-button>
   <div ng-transclude id="widget-removal"></div>
-<!--   <md-button class="widget-action widget-remove md-icon-button" -->
-<!--              aria-label="remove {{ widget.title }} widget from your home screen" -->
-<!--              ng-click="layoutCtrl.removePortlet(widget.nodeId, widget.title)" -->
-<!--              ng-hide="GuestMode || cantRemove"> -->
-<!--     <md-icon>close</md-icon> -->
-<!--   </md-button> -->
-
   <a aria-labelledby="appTitle_widget.title-{{::widget.nodeId}}" tabindex="0" ng-href="{{widget.url}}" target="{{::widget.target}}">
     <div class="icon-container">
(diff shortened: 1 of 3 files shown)


uw-frame-components/portal/widgets/partials/compact-widget-card.html
```diff
@@ -0,0 +1,24 @@
...
+    </md-tooltip>
+    <md-icon>info</md-icon>
+  </md-button>
+  <div ng-transclude id="widget-removal"></div>
+<!--   <md-button class="widget-action widget-remove md-icon-button" -->
+<!--              aria-label="remove {{ widget.title }} widget from your home screen" -->
+<!--              ng-click="layoutCtrl.removePortlet(widget.nodeId, widget.title)" -->
+<!--              ng-hide="GuestMode || cantRemove"> -->
+<!--     <md-icon>close</md-icon> -->
+<!--   </md-button> -->
```
ChristianMurphy: Is the commented out code needed?

vertein: Great catch!  Removed it to prove that I didn't need it, but didn't actually remove it

uw-frame-components/portal/widgets/partials/compact-widget-card.html
```diff
@@ -0,0 +1,17 @@
+<md-card class="list-content" id="widget-id-{{::widget.nodeId}}">
+  <md-button class="widget-action widget-info md-icon-button" aria-label="see a brief description of this app">
```
apetro: `aria-label` is good. As I understand it it would be better if the label included the name of the app that one would see a brief description of, so that screen reader user has more context in consuming that label.

vertein: Good catch.  I just copied and pasted a partial from before.  It's always good to be better.  Updated.

PR #419, opened 2017-04-28
Adds Error message to unauthorized widget.

Doug-Reed:
Adds error text to unauthorized widget appearing in layout. 

----

- [X ] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

### DIFF
--- a/uw-frame-components/portal/widgets/services.js
+++ b/uw-frame-components/portal/widgets/services.js
@@ -19,10 +19,19 @@ define(['angular'], function(angular) {
           }
         })
         .catch(function(error) {
+          return getErrorPage(fname);
           $log.warn('Error getting marketplace entry for ' + fname);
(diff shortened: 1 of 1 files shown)


uw-frame-components/portal/widgets/services.js
```diff
@@ -19,11 +19,20 @@ define(['angular'], function(angular) {
           }
         })
         .catch(function(error) {
+          return getErrorPage(fname);
```
apetro: Does this `return` placement before the two following `$log.` statements mean those log statements are never called?

Doug-Reed: Probably. I'll reorder.

PR #420, opened 2017-04-28
Beautify permission-denied widget error

thevoiceofzeke:
#419 added the ability to show an error message when a user requests a widget they don't have permission to use/see. This PR just leverages some existing functionality (maintenance mode) to display the error in a similarly attractive way.

### Screenshot
screen shot 2017-04-28 at 2 08 21 pm

### DIFF
--- a/uw-frame-components/portal/widgets/services.js
+++ b/uw-frame-components/portal/widgets/services.js
@@ -1,10 +1,10 @@
 'use strict';
 
-define(['angular'], function(angular) {
+define(['angular'], function (angular) {
 
   var app = angular.module('portal.widgets.services', []);
 
-  app.factory('widgetService', ['$http', '$log', 'SERVICE_LOC', function($http, $log, SERVICE_LOC) {
+  app.factory('widgetService', ['$http', '$log', 'SERVICE_LOC', function ($http, $log, SERVICE_LOC) {
 
     /**
      * Get the a single app's full entity file as JSON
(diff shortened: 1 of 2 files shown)


uw-frame-components/portal/widgets/services.js
```diff
@@ -13,35 +13,49 @@ define(['angular'], function(angular) {
...
+        title: fname,
+        mdIcon: "help",
+        widgetType: "generic",
+        widgetConfig: {
+          additionalText: "Please contact your help desk if you feel you should be able to access this content"
+        },
+        widgetTemplate: "<div class='overlay__maintenance-mode'>"
+          + "<div class='maintenance-content'>"
+          + "<p><md-icon class='md-warn'>warning</md-icon></p>"
+          + "<p>You do not have permission to access this content. If you feel this is an error, please contact your help desk.</p>"
```
apetro: Is lack of permission the only way to catch an error at line 21? What happens if the `fname` doesn't map to anything? The underlying portlet-definition is expired? MyUW's web services are unavailable because the `/portal` context collapsed under its own bloated weight?

Opportunity to differentiate amongst failure modalities, or in the meantime to use more generic error messaging for this default "something bad happened but dunno what" case?

Doug-Reed: Do we want the user to be able to differentiate? Does it really matter from a user's perspective? The only time they'll see this is if they have something hardcoded onto a layout, and it doesn't work. I think the generic message is probably the way to go. 

What's a concise way to say "Your widget is FUBAR. Call for help."?

apetro: "Oops. MyUW couldn't display this widget. If this continues and bothers you enough, please _contact the Help Desk_."

PR #421, opened 2017-05-01
cleaning up filesystem errors on build-static

davidmsibley:
changes to frame weren't showing up when I ran static:dev, so I went through and fixed the build script.  

Standardized error throwing, copy now overwrites.

----

### DIFF
--- a/tools/uw-frame-static/build.js
+++ b/tools/uw-frame-static/build.js
@@ -12,9 +12,12 @@ var exec_handler = function(error, stdout, stderr) {
 
 var mkdirp = require('mkdirp');
 var copy = require('recursive-copy');
+var copyOptions = {
+  overwrite: true
+}
 
-mkdirp('uw-frame-static/target/css/themes/', function (err) {
-  if (err) throw error;
+mkdirp('uw-frame-static/target/css/themes/', function (error) {
+  if (error) throw error;
 });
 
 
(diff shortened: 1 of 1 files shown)


tools/uw-frame-static/build.js
```diff
@@ -12,9 +12,12 @@ var exec_handler = function(error, stdout, stderr) {
...
 var mkdirp = require('mkdirp');
 var copy = require('recursive-copy');
+var copyOptions = {
+  overwrite: true
+}
 
-mkdirp('uw-frame-static/target/css/themes/', function (err) {
-  if (err) throw error;
+mkdirp('uw-frame-static/target/css/themes/', function (error) {
+  if (error) throw error;
```
ChristianMurphy: could `error` be confused with `Error` here?

tools/uw-frame-static/build.js
```diff
@@ -12,9 +12,12 @@ var exec_handler = function(error, stdout, stderr) {
 
 var mkdirp = require('mkdirp');
 var copy = require('recursive-copy');
+var copyOptions = {
```
ChristianMurphy: assuming Node.js 4+ is the target run time, should this be a `const`?

tools/uw-frame-static/build.js
davidmsibley: I want to trust that `Error` would be immediately noticeable due to classes starting with capitals by convention.  If someone saw `Error` and not `Error()` or `new Error()` that should be picked up during review.

tools/uw-frame-static/build.js
davidmsibley: yup. and we're using const elsewhere in the file.  I'll update the file to use it consistently.

tools/uw-frame-static/build.js
ChristianMurphy: Got it, you are expecting Google code style's [`new-cap`](http://eslint.org/docs/rules/new-cap) lint rule to catch this.

PR #429, opened 2017-05-04
Fixing trivial ESLint checks

davidmsibley:
Fixing the grab bag of style misgivings, and pulling in the changes from master.

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
   "eslintConfig": {
     "root": true,
     "parserOptions": {
-      "ecmaVersion": 6
+      "ecmaVersion": 5
     },
     "env": {
       "browser": true,
@@ -144,7 +144,7 @@
       "max-len": "warn",
       "promise/catch-or-return": "warn",
       "promise/always-return": "warn",
-      
+
       "compat/compat": "error",
       "arrow-parens": "off",
       "constructor-super": "off",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "uw-frame",
-  "version": "4.0.3",
+  "version": "4.1.0",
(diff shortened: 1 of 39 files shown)


package.json
```diff
@@ -134,6 +134,17 @@
...
+      "angular/module-setter": "warn",
+      "angular/module-getter": "warn",
+      "angular/typecheck-object": "warn",
+      "angular/typecheck-array": "warn",
+      "angular/typecheck-string": "warn",
+      "angular/no-private-call": "warn",
+      "max-len": "warn",
+      "promise/catch-or-return": "warn",
+      "promise/always-return": "warn",
+
```
davidmsibley: These are the list of rules that have the most violations.  With the misc violations fixed, we can go through one by one to fix these.

package.json
```diff
@@ -104,7 +104,7 @@
   "eslintConfig": {
     "root": true,
     "parserOptions": {
-      "ecmaVersion": 5
+      "ecmaVersion": 6
```
ChristianMurphy: What was causing errors parsing with the ES5 standard?

davidmsibley: almost everything is ES5 _except_ [tools/uw-frame-static/build.js](https://github.com/UW-Madison-DoIT/uw-frame/blob/master/tools/uw-frame-static/build.js)

davidmsibley: and unfortunately I can't change versions of the parser per file, it's a project level thing.

ChristianMurphy: Parser can be changed per folder.
an `.eslintrc.yml` or `eslintrc.json` file in a folder will override project level configuration.

ChristianMurphy: also configurations are merged, so only the `parserOptions` property would need to be set.

PR #432, opened 2017-05-04
Eslint promise rule fixes

davidmsibley:
fixes `promise/always-return` and `promise/catch-or-return`

----


davidmsibley:
hah. I forgot I did that.  I recall @vertein de-spaghetti'd a promise a month or so ago, I could probably look that up as an example and take a shot.

PR #433, opened 2017-05-04
"fix" eslint angular/no-private-call

davidmsibley:
RFP for a way to rewrite this function without the `next.$$route` call.

----

### DIFF
--- a/uw-frame-components/portal/main.js
+++ b/uw-frame-components/portal/main.js
@@ -325,11 +325,11 @@ define([
         // https://github.com/Gillespie59/eslint-plugin-angular/issues/231
         // eslint-disable-next-line angular/on-watch
         $rootScope.$on('$routeChangeStart', function(event, next, current) {
-          var searchValue = '';
-          var paramToTackOn = '';
-          if(next.$$route && next.$$route.searchParam) {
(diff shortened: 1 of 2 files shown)


uw-frame-components/portal/main.js
```diff
@@ -325,11 +325,11 @@ define([
         // https://github.com/Gillespie59/eslint-plugin-angular/issues/231
         // eslint-disable-next-line angular/on-watch
         $rootScope.$on('$routeChangeStart', function(event, next, current) {
-          var searchValue = '';
-          var paramToTackOn = '';
-          if(next.$$route && next.$$route.searchParam) {
-            paramToTackOn = APP_FLAGS.gaSearchParam || 'q';
-            searchValue = next.params[next.$$route.searchParam];
+          // eslint-disable-next-line angular/no-private-call
+          var nextRoute = next.$$route;
```
ChristianMurphy: Isn't next of type Route? Why is the internal route object needed?
https://docs.angularjs.org/api/ngRoute/service/$route#$routeChangeStart

davidmsibley: you're right, searchParam is accessible just via `next.searchParam`

PR #442, opened 2017-05-09
fixing errors and not returning reasons

davidmsibley:
rejecting promises rather than throwing errors.

----

### DIFF
--- a/uw-frame-components/portal/features/services.js
+++ b/uw-frame-components/portal/features/services.js
@@ -1,6 +1,5 @@
 'use strict';
 
-/* this is the spaghetti'est thing I've ever seen. I'm not touching this. */
 /* eslint-disable promise/always-return */
 
 define(['angular'], function(angular) {
@@ -84,7 +83,7 @@ define(['angular'], function(angular) {
         .getValue('lastviewedannouncementid')
         .then(function(data) {
           if (!data || !data.id) {
-            throw Error('no legacy seenAnnouncements');
+            return $q.reject('no legacy seenAnnouncements');
           }
           // create an array with the seen ids
           var seenAnnouncements = [];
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "uw-frame",
-  "version": "4.1.0",
+  "version": "4.1.1",
```
vertein: ?

davidmsibley: whoops, disregard.  I was messing around with my local build, I don't know why I changed this.

davidmsibley:
It does handle the promise correctly on throw, I just went this route because the error bubbles up and you get a stack trace in the console. It seemed overly ugly for a pretty inconsequential failure.

ChristianMurphy:
> It does handle the promise correctly on throw, I just went this route because the error bubbles up and you get a stack trace in the console. It seemed overly ugly for a pretty inconsequential failure.

I'm not quite sure I follow what you are saying.
Let me rephrase to what I interpret that as:

> `throw`/`Promise` handles the exception as expected.
> However the `Error` is causing the resulting value to bubble up into a `console.error`.
> This is a handler for legacy code and should not be shown as a system/application exception.

Is that an accurate understanding? :thought_balloon:

davidmsibley:
pretty much.  I don't like errors propagating to the console if they're really just part of the normal control flow.  But that's just superficial preference by me, if it's a big deal, we can go back to throw.

ChristianMurphy:
`throw` or `$q.resolve` could work equally well. :man_shrugging: 
I would lean toward `throw` only because it better aligns with lint rule. :white_check_mark: 

>  I don't like errors propagating to the console if they're really just part of the normal control flow. 

I hear ya on that. :+1: 
`async await` in ES8 will give some really clean ways to handle early exit.
In the mean time [custom error types](http://stackoverflow.com/a/5251506) could allow to customized when logs are shown and at what level.

Or lowering the log level works as well. :slightly_smiling_face:

ChristianMurphy:
aside: [Coveralls not reporting status is an issue with coveralls](http://status.coveralls.io/incidents/ynqhpvsbrd96) not with this PR.

PR #443, opened 2017-05-09
pulling string into configuration

davidmsibley:
if we pull the '.json' suffix into configuration, we're able to do some creative stuff with the uri we call for widget data (like using a base64 data uri).

----


davidmsibley:
would be used here: https://github.com/thevoiceofzeke/widget-creator/pull/2

PR #448, opened 2017-05-23
Improve usability for screen readers

thevoiceofzeke:
Addresses [MUMUP-2920](https://jira.doit.wisc.edu/jira/browse/MUMUP-2920)

As reported by Sandi (UX):

> There is an issue where the app loads later than the frame and the screen reader reads the frame content and skips the app content - only reading the footer.

**In this PR**:
- Widget titles are now read before their descriptions (and only one time). This should give screen reader users a better sense of context and hierarchy.
- Added tabindex attributes to improve keyboard navigation and signify focusability
- Changed app-header role attribute to reflect what it really is
- Added ~aria-live~ attribute to main content area (to signify dynamic content)
- *(tangential)* Slightly lengthened time of splash screen fade to reduce exposure of pre-CSS content on slower connections (in lieu of a real fix)
- *(tangential)* Added a text filter to make the theme selector options a little prettier
- *(tangential)* Replaced some annoying single quotes in HTML

Sandi tested these changes in predev environment using VoiceOver and she has indicated that the bug seems to be fixed. This is marked WIP because there is more testing to do (in NVDA and JAWS), as well as the issue that VoiceOver in more recent versions of OS X seems to have its own set of bugs that are not remedied by these changes.

----

### DIFF
--- a/uw-frame-components/portal/settings/controllers.js
+++ b/uw-frame-components/portal/settings/controllers.js
@@ -3,6 +3,21 @@
 define(['angular'], function(angular) {
   return angular.module('portal.settings.controllers', [])
 
+    .filter('prettyName',function() {
+      return function(input) {
+        if (input) {
+          if (input.indexOf('uw-') > -1) {
+            var partOne = input.slice(0, 3).toUpperCase();
+            var partTwo = input.slice(input.indexOf('-') + 1)
+              .replace(/-/g, ' ')
+              .replace(/\w\S*/g, function(txt){return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();});
+            return partOne + partTwo;
+          } else {
+            return input;
+          }
+        }
+      }
+    })
 
   .controller('PortalBetaSettingsController', [
     '$sessionStorage', '$scope', '$mdTheming', 'portalSkinService',
@@ -74,4 +89,5 @@ define(['angular'], function(angular) {
 
       init();
     }]);
+
  });
--- a/uw-frame-components/portal/settings/controllers.js
+++ b/uw-frame-components/portal/settings/controllers.js
@@ -3,7 +3,10 @@
 define(['angular'], function(angular) {
   return angular.module('portal.settings.controllers', [])
 
-    .filter('prettyName',function() {
+  /**
+   * For UW-prefixed theme names, formats the name for a more attractive display in the theme selection settings box.
+   */
+    .filter('formatThemeNameForDisplay',function() {
       return function(input) {
         if (input) {
           if (input.indexOf('uw-') > -1) {
--- a/uw-frame-components/portal/settings/controllers.js
+++ b/uw-frame-components/portal/settings/controllers.js
@@ -4,22 +4,26 @@ define(['angular'], function(angular) {
   return angular.module('portal.settings.controllers', [])
 
   /**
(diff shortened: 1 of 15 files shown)


uw-frame-components/static.html
```diff
@@ -9,8 +9,8 @@
   }
 
   .slowfade {
```
davidmsibley: sloooow fade. I don't think we want to make this longer to try to solve the pre-css problem. (Or at least we should use a different easing than linear: http://easings.net/ ) This slow fade is the reason for that period of unresponsiveness on page load that I've heard @apetro grumble about :)

uw-frame-components/portal/settings/controllers.js
```diff
@@ -3,6 +3,21 @@
 define(['angular'], function(angular) {
   return angular.module('portal.settings.controllers', [])
 
+    .filter('prettyName',function() {
```
apetro: Suggest a comment hinting at what this filter does. Suggest re-name to better reflect what the filter does or is intended to do.  I think this converts ugly technical theme identifiers into attractive user-facing names for themes.

uw-frame-components/portal/widgets/partials/widget-card.html
```diff
@@ -23,7 +23,7 @@
     </div>
 
     <md-card-header-text>
-      <span class="md-title" style="text-align: center;" aria-label="{{ widget.title }}" tabindex="0">
+      <span class="md-title" style="text-align: center;" aria-label="{{ widget.title }}" aria-hidden="true">
```
apetro: Is it weird to `aria-label` an element that's also `aria-hidden`?
Does this mean that if the widget is in `MAINTENANCE` lifecycle state then screen readers won't read the title of the widget at all? 

I think this change is happening to make the read-first description more sensible in the context of the widget. But probably usually users don't want to read the description at all?

Consider instead aria-hiding the info md-button, not hiding this md-title span, and adding a link to the details page, visible only to screen readers, with the app description included as the description of what that link is going to take you to? Disclaimer: haven't tested this hypothetical alternative, so maybe its accessibility is bad.

uw-frame-components/static.html
thevoiceofzeke: Eh, I'll remove it increase. I don't believe it has anything to do with the lag in responsiveness though. That seems to be due to something going on in the angularJS bootstrapping process.

uw-frame-components/portal/widgets/partials/widget-card.html
thevoiceofzeke: For now, I think just not hiding the title would be a good option.

uw-frame-components/portal/settings/partials/settings.html
```diff
@@ -15,7 +15,7 @@
         <h4 class="md-subhead">Theme Selector</h4>
         <md-input-container>
           <md-select ng-model="portal.theme" placeholder="Pick which place you call home" aria-label="Select a theme">
-            <md-option ng-repeat="item in THEMES" ng-value="item">{{ item.name }}</md-option>
+            <md-option ng-repeat="item in THEMES" ng-value="item">{{ item.name | prettyName }}</md-option>
```
apetro: Needs to be `| formatThemeNameForDisplay` rather than `| prettyName`?

PR #449, opened 2017-05-25
Fixes skinning.md lint errors

vertein:
Fixes linting errors.  If we mark code as javascript, linting picks it up and makes sure it's valid.

Automating linting been failing for a day or so

----

Contributor License Agreement adherence:

- [x] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

[Apereo CLA roster]: http://licensing.apereo.org/completed-clas
[Contributor License Agreements]: https://www.apereo.org/licensing/agreements

PR #450, opened 2017-05-25
Announcement bug fix

vertein:
Fixes a bug when the code updates legacy dismissed announcement id storage to the new format.  Bug introduced about a month ago where if there is not any legacy dismissal ids, a chain reaction of rejection causes no announcements of any type to show.  This pr also adds some return type documentation, but doesn't add unit tests to prevent this in the future.

Marked WIP because of the failing lint test that https://github.com/UW-Madison-DoIT/uw-frame/pull/449 takes care of.  When that is merged, I'll rebase, then this should be good to go and will pass the CI

----

### DIFF
--- a/uw-frame-components/portal/features/services.js
+++ b/uw-frame-components/portal/features/services.js
@@ -73,17 +73,18 @@ define(['angular'], function(angular) {
      * If keyValueService is Active and there exists legacy announcement storage
      * this will convert that to the new store all seen announcement ids rather
      * than just the latest.  Will then delete the legacy announcement storage
+     * @return {String[]} ids that were updated from legacy storage system
      */
     var updateLegacySeenAnnouncements = function() {
       if (!keyValueService.isKVStoreActivated()) {
-        return $q.resolve(null);
+        return $q.resolve([]);
       }
 
       return keyValueService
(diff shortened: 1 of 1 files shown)


uw-frame-components/portal/features/services.js
```diff
@@ -275,6 +279,8 @@ define(['angular'], function(angular) {
             .filter(filterSeenPopups)
             .filter(filterExpiredPopups)
             .filter(filterUnEnabledPopups);
+        }else{
```
ChristianMurphy: :art: could the else have spaces added around it?

uw-frame-components/portal/features/services.js
```diff
@@ -235,6 +237,8 @@ define(['angular'], function(angular) {
             }
           };
           return announcements.filter(hasNotSeen);
+        }else{
```
ChristianMurphy: :art: could the else have spaces added around it?

vertein: Updated. Thanks!

uw-frame-components/portal/features/services.js
vertein: Updated. Thanks!

PR #451, opened 2017-06-02
Frame works withu portal master

vertein:
Getting frame to work with uPortal master.  This should take care of most of the big things.

* Adds a default theme to go with uPortal
* Fixes a bug with launch urls that UW-Madison never sees because of our explicit handling of urls in portlet definition files.
* Fixes a bug where widget url wasn't working (again, never saw it because of our explicit handling of icons in portlet definition files).

----

### DIFF
--- a/tools/uw-frame-static/build.js
+++ b/tools/uw-frame-static/build.js
@@ -18,7 +18,7 @@ mkdirp('uw-frame-static/target/css/themes/', function(error) {
 });
 
 
-const themes = ['uw-madison', 'uw-system', 'uw-river-falls',
+const themes = ['default', 'uw-madison', 'uw-system', 'uw-river-falls',
   'uw-stevens-point', 'uw-milwaukee', 'uw-whitewater', 'uw-stout',
   'uw-superior', 'uw-platteville', 'uw-parkside', 'uw-oshkosh',
   'uw-greenbay', 'uw-lacrosse', 'uw-eau-claire', 'uw-extension',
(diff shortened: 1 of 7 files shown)


Doug-Reed:
:+1:

uw-frame-java/pom.xml
```diff
@@ -235,6 +235,18 @@
...
         <executions>
           <execution>
+            <id>rename-default-file</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>rename</goal>
+            </goals>
+            <configuration>
+              <sourceFile>${basedir}/target/frame/css/themes/default.css</sourceFile>
+              <destinationFile>${basedir}/target/frame/css/themes/default.${project.version}.css</destinationFile>
```
davidmsibley: cool cache breaking, but how does it get pulled in to the page?

vertein: https://github.com/UW-Madison-DoIT/uw-frame/blob/master/uw-frame-java/src/main/webapp/frame.jsp#L5

vertein: "cool"

tools/uw-frame-static/build.js
```diff
@@ -18,7 +18,7 @@ mkdirp('uw-frame-static/target/css/themes/', function(error) {
 });
 
 
-const themes = ['uw-madison', 'uw-system', 'uw-river-falls',
+const themes = ['default', 'uw-madison', 'uw-system', 'uw-river-falls',
```
apetro: I checked how theme preference is stored client-side in session storage: it's by name rather than by array index (hurray!) so this should deploy cleanly even if static theming mattered in prod, which I bet we're not actually doing any of anyway. :smile:

PR #453, opened 2017-06-03
Enables notifications to work in static frame

vertein:
----

### DIFF
--- a/uw-frame-components/js/app-config.js
+++ b/uw-frame-components/js/app-config.js
@@ -11,10 +11,10 @@ define(['angular'], function(angular) {
         })
         .value('SERVICE_LOC', {
             'aboutURL': null,
-            'groupURL': '/portal/api/groups',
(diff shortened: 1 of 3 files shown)


uw-frame-components/js/app-config.js
```diff
@@ -11,10 +11,10 @@ define(['angular'], function(angular) {
         })
         .value('SERVICE_LOC', {
             'aboutURL': null,
-            'groupURL': '/portal/api/groups',
+            'groupURL': 'staticFeeds/groups.json',
```
apetro: Concern: is this the default value as experienced by frame-based apps? Such that frame-based apps not overriding this value will stop correctly grabbing groups from `/portal/api/groups` unless they start overriding this value in app-local config?

vertein: Good heads up.

PR #457, opened 2017-06-12
Reverts groups url back to previous default state

vertein:
Reverts this change, so that default behavior won't change in the next release.

Heads up due to this [comment](https://github.com/UW-Madison-DoIT/uw-frame/pull/453/files/e3ab440dc4bdf6083108195eda045dcb7ab9f837#diff-de864f63fac068677f42dde719b42604R14)

Will need to think about default behaviors as a whole later on (soon)...

----

### DIFF
--- a/uw-frame-components/js/app-config.js
+++ b/uw-frame-components/js/app-config.js
@@ -11,10 +11,10 @@ define(['angular'], function(angular) {
         })
         .value('SERVICE_LOC', {
             'aboutURL': null,
-            'groupURL': 'staticFeeds/groups.json',
+            'groupURL': '/portal/api/groups',
             'kvURL': '/storage',
             'loginSilentURL': '/portal/Login?silent=true',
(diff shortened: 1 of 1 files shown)


uw-frame-components/js/app-config.js
```diff
@@ -14,7 +14,7 @@ define(['angular'], function(angular) {
             'groupURL': 'staticFeeds/groups.json',
             'kvURL': '/storage',
             'loginSilentURL': '/portal/Login?silent=true',
-            'notificationsURL': 'staticFeeds/notifications.json',
```
davidmsibley: Did you mean to change the notifications url and not the groups url on line 14?

vertein: You are correct!  Will reply when corrected.

vertein: Thank you for catching that.  Updated.

PR #467, opened 2017-07-05
B opt out photo

Doug-Reed:
This changeset provides the functionality to allow a user to opt out of using their photo as an avatar. (The photo functionality isn't built yet.)

If a user is in a theme which allows an avatar, (as specified by putting an avatar value in uw-frame-components/js/frame-config.js), and the user has toggled yes to "Allow use of my photo" in /settings, then that user will currently see the word "PIC". This is a placeholder until the photo functionality is built. 

The user can toggle off the switch and opt out. 
----

- [ X] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].


uw-frame-components/js/frame-config.js
```diff
@@ -1069,5 +1070,11 @@ define(['angular'], function(angular) {
             'description': 'This flag disables announcement group filtering ' +
               'if you have it enabled (page refresh required)',
           },
+          {
+            'id': 'optInAvatar',
+            'title': 'Allow Use Of My Photo',
+            'description': 'This flag allows use of your photograph ' +
+              'if you have it enabled (page refresh required)',
+          },
```
davidmsibley: This method of opting in would require the user to know about and navigate to our nuts and bolts settings page. We should keep this feature flag, but maybe add a toggle in the user settings area?

apetro: Agree. Would be best for the user-facing opt-in / opt-out setting to be in the `/user-settings` page. Could hide the presentation of that setting behind its own feature flag, but probably better to just hide it behind whether the active theme supports avatar image feature.

Doug-Reed:
I'll move that toggle setting.

davidmsibley:
Can this be closed since #469 was merged?

PR #468, opened 2017-07-06
error checking weather data response

davidmsibley:
The weather widget will not put itself in an error state unless a network error occurs. This checks the data response for reported errors, and treats that as an error condition.

----

### DIFF
--- a/uw-frame-components/portal/widgets/controllers.js
+++ b/uw-frame-components/portal/widgets/controllers.js
@@ -513,7 +513,10 @@ define(['angular'], function(angular) {
         $scope.loading = false;
 
         // If we got some data, populate widget content
-        if (data) {
+        if (
+          data && 2 === data.length && data[0] && data[1] &&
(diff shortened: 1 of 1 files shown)


uw-frame-components/portal/widgets/controllers.js
```diff
@@ -513,7 +513,10 @@ define(['angular'], function(angular) {
         $scope.loading = false;
 
         // If we got some data, populate widget content
-        if (data) {
+        if (
+          data && 2 === data.length && data[0] && data[1] &&
+          (!data[0].errors || (data[0].errors && 0 >= data[0].errors.length))
```
ChristianMurphy: This is an area where JSON Schema could come in handy.

```json
{
  "type": "array",
  "minItems": 2,
  "maxItems": 2,
  "items": {
    "type": "object",
    "properties": {
      "errors": {
        "type": "array",
        "maxItems": 0
      }
    }
}
```

PR #473, opened 2017-07-20
Add URL capability to list of links type

Doug-Reed:
This change adds functionality to the list-of-links widget type to retrieve the list from a URL rather than a hardcoded portlet preference. 

----

- [ x] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

### DIFF
--- a/docs/make-a-widget.md
+++ b/docs/make-a-widget.md
@@ -138,6 +138,13 @@ Follow these steps for each of the predefined widget types described in this doc
   </value>
 </portlet-preference>
 
+#### Configuring your list of links from a URL
+
+* Rather than hardcode your links into a portlet preference, you can point to a URL and get your links dynamically. 
+1. Omit the "links" entry in the widgetConfig JSON.
+2. In widgetConfig, add the following:
+   "getLinksURL": "true"
+3. Point the widget's widgetURL property to the location of the feed. 
 ```
 
 #### Additional information
--- a/docs/make-a-widget.md
+++ b/docs/make-a-widget.md
@@ -137,6 +137,7 @@ Follow these steps for each of the predefined widget types described in this doc
     }]]>
   </value>
 </portlet-preference>
+```
 
 #### Configuring your list of links from a URL
 
@@ -145,7 +146,7 @@ Follow these steps for each of the predefined widget types described in this doc
 2. In widgetConfig, add the following:
    "getLinksURL": "true"
 3. Point the widget's widgetURL property to the location of the feed. 
-```
+
 
 #### Additional information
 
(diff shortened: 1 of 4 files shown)


docs/make-a-widget.md
```diff
@@ -138,6 +138,13 @@ Follow these steps for each of the predefined widget types described in this doc
   </value>
 </portlet-preference>
 
+#### Configuring your list of links from a URL
+
+* Rather than hardcode your links into a portlet preference, you can point to a URL and get your links dynamically. 
+1. Omit the "links" entry in the widgetConfig JSON.
+2. In widgetConfig, add the following:
+   "getLinksURL": "true"
```
apetro: Feasible to make setting the `widgetURL` have this configuration effect, rather than also introducing `getLinksURL`? Configuration that can't be ambiguous is nice -- with the setting overlap, one could set `widgetURL` and have it have no effect because overlooked setting `getLinksURL`.

docs/make-a-widget.md
```diff
@@ -138,6 +138,13 @@ Follow these steps for each of the predefined widget types described in this doc
   </value>
 </portlet-preference>
 
+#### Configuring your list of links from a URL
```
ChristianMurphy: Could this content be moved out of the code block that ends on line 148?

uw-frame-components/portal/widgets/controllers.js
```diff
@@ -24,6 +24,18 @@ define(['angular'], function(angular) {
         case 'list-of-links':
           // If the list of links only has one link and it's the
           // same as the launch button url, display a basic widget
+          if (widget.widgetConfig.getLinksURL) {
```
apetro: `if (widget.widgetURL) {`

uw-frame-components/portal/widgets/controllers.js
```diff
@@ -24,6 +24,18 @@ define(['angular'], function(angular) {
         case 'list-of-links':
           // If the list of links only has one link and it's the
```
apetro: This comment would be better bumped down to document
```
if (widget.widgetConfig.links.length === 1 && widget.altMaxUrl &&
                widget.widgetConfig.links[0].href === widget.url) {
```

docs/make-a-widget.md
Doug-Reed: What happens in the case where we have a valid widgetURL and also links in the widgetConfig?

apetro: It's a good question that would need an unambiguous answer. Offhand, I'd say in that case it should try to use the remote content and fall back on the local `widgetConfig` content in lieu of an error experience when the remote call fails.

PR #489, opened 2017-08-11
Acknowledge bundled dependencies and add NOTICE file

apetro:
Trying to follow Apache Software Foundation practices.

So, uPortal-application-framework has two third party things bundled in the source code (glyphicons and sortable.js), both via The MIT License. [Per ASF guidance](http://www.apache.org/dev/licensing-howto.html#permissive-deps), acknowledged these at the bottom of `LICENSE`. (It was news to me to prefer to do this in `LICENSE` rather than `NOTICE`).

uPortal-application-framework has lots of other dependencies, but doesn't bundle their bits, so need not acknowledge them formally in `NOTICE` or `LICENSE`.

In `NOTICE`, I don't actually understand why it would be correct to say it's "Copyright the Apereo Foundation" since the authors rather than the foundation retain actual copyright ownership -- contributors only license their contributions to Apereo and to everyone downstream. But that's [the ASF guidance](http://www.apache.org/dev/licensing-howto.html#simple), so did that.

----

Special thanks to @ChristianMurphy for removing un-needed `less-1.6.2.js` in #439 which saves figuring out how to acknowledge that no-longer-bundled dependency.

----

I am not a lawyer. This is not legal advice. Just a layperson trying to muddle through license compliance.

----

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -200,3 +200,30 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 
+----
+
+This product bundles Glyphicon Halflings ( http://glyphicons.com/
+ ) via Bootstrap 3's distributing these under The MIT License.
+
+This product bundles angular-ui / ui-sortable / sortable.js ( https://github.com/angular-ui/ui-sortable/blob/master/src/sortable.js ) under The MIT License.
+
+
+The MIT License (MIT)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -200,3 +200,30 @@
...
 
+----
+
+This product bundles Glyphicon Halflings ( http://glyphicons.com/
+ ) via Bootstrap 3's distributing these under The MIT License.
+
+This product bundles angular-ui / ui-sortable / sortable.js ( https://github.com/angular-ui/ui-sortable/blob/master/src/sortable.js ) under The MIT License.
+
+
+The MIT License (MIT)
```
ChristianMurphy: Per the Apache guide, MIT license should not not appended.

> In LICENSE, add a pointer to the dependency's license within the distribution and a short note summarizing its licensing

ref: https://www.apache.org/dev/licensing-howto.html#permissive-deps
ref: https://mail-archives.apache.org/mod_mbox/incubator-general/201203.mbox/%3CCBFE3722-1B5C-48FC-AD82-0931EEC6F143@gbiv.com%3E

ChristianMurphy: > NOTE: It's also possible to include the text of the 3rd party license within the LICENSE file. This is best reserved for short licenses. It's important to specify the version of the dependency as licenses are sometimes changed.

However

> Pointers are sufficient.

NOTICE
```diff
@@ -0,0 +1,6 @@
+Apereo uPortal-application-framework aka uw-frame
```
ChristianMurphy: I find it odd that there is a `NOTICE` file to say that there are no notices.
But it seems to be an ASF recommendation so :+1:

LICENSE
apetro: > add a pointer to the dependency's license within the distribution

Note _within the distribution_. If we don't have a full copy of the (short) MIT License directly in LICENSE, then we'll have to have a full copy of it somewhere else.

ChristianMurphy: Would a link to https://opensource.org/licenses/MIT
Or reference links like
1. node_modules/bootstrap/LICENSE
2. node_modules/font-awesome/LICENSE
3. node_modules/normalize.less/LICENSE

work?

apetro: I don't believe references to `node_modules` or URLs are sufficient, no. If we're distributing the bits of the third party software, we have to also distribute the bits of the third party software's license.

`node_modules` aren't in the distribution as such -- they're stuff node pulls down in resolving pointers in package metadata. It's subtle, but the point is those copies of the LICENSEs aren't actually in our "distribution", in the thing we're distributing.

Likewise, a link doesn't fulfill the MIT license requirement that 

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

!L. Not legal advice.

ChristianMurphy: hmm :thinking: 
well better safe than sorry :+1:
thanks for looking into that @apetro!

PR #492, opened 2017-08-15
fix: Adds npm dependencies to right area MUMUP-3004

vertein:
Tested out the build ignoring the dev dependencies.  Found out that 4 more dependencies are actually required for the build to succeed.  This fix adds them there.

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
   },
   "homepage": "https://github.com/UW-Madison-DoIT/uw-frame",
   "devDependencies": {
-    "autoprefixer": "^7.1.2",
     "concurrently": "^3.1.0",
     "eslint": "^4.2.0",
     "eslint-config-angular": "^0.5.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -105,7 +101,11 @@
   "dependencies": {
     "bootstrap": "^3.3.6",
     "font-awesome": "^4.6.3",
-    "normalize.less": "^1.0.0"
+    "normalize.less": "^1.0.0",
+    "less": "2.7.2",
+    "autoprefixer": "^7.1.2",
```
ChristianMurphy: Please alphabetize dependencies, or run `npm install --save less` to get npm to automatically sort.

---

Also regenerate `package-lock.json`.
npm v5+ pulls dependencies purely based off the lockfile, if the lock is out of sync with the package, the changes here would not appear.

vertein:
Cool.  We're running into a problem with a dev dependency that is failing to install on our build server because our build server doesn't have a c compiler currently installed.  I'll look into a different way to solve that.  Thanks for the info!

ChristianMurphy:
@vertein `optionalDependencies` would be one option.
If the C/C++ code need to be built, let me know what the dependency is and I can look into updating Travis and AppVeyor to support native API builds.

PR #499, opened 2017-08-22
Animate priority notifications [nice-to-have] 

thevoiceofzeke:
[MUMUP-3006](https://jira.doit.wisc.edu/jira/browse/MUMUP-3006): "As a user, I'd like a smoother experience when dismissing priority notifications, so that I can see the action happening, rather than just a disappearing bar."

**In this PR**:
- Added animations for appearance/dismissal of priority notifications
- Adjusted design of priority notifications to be more attractive (hopefully)
- Made a little more room for options buttons in app-header
- Adjusted some z-indexes so not to overlay Angular Material modals
- Adjusted confusing CSS dealing with fixed-positioned elements
- Username in user menu no longer all caps, first letter capitalized instead
- Added LESS variables for 5 levels of depth (per material design)
- Renamed some LESS variables so not to tie them to specific HTML elements

### Demo
notifications-demo

#### Priority notifications
screen shot 2017-08-22 at 1 25 08 pm

#### Modal depth fix
screen shot 2017-08-22 at 10 58 44 am

#### Username menu
screen shot 2017-08-22 at 10 21 07 am

----

### DIFF
--- a/uw-frame-components/portal/main/controllers.js
+++ b/uw-frame-components/portal/main/controllers.js
@@ -109,7 +109,7 @@ define(['angular', 'require'], function(angular, require) {
 
       if (vm.user.firstName || vm.user.displayName) {
         vm.username = vm.user.firstName ?
-          vm.user.firstName : vm.user.displayName;
+          vm.user.firstName.toLowerCase() : vm.user.displayName.toLowerCase();
       }
       vm.firstLetter = vm.username.substring(0, 1);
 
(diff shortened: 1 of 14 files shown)


uw-frame-components/portal/main/controllers.js
```diff
@@ -109,7 +109,7 @@ define(['angular', 'require'], function(angular, require) {
 
       if (vm.user.firstName || vm.user.displayName) {
         vm.username = vm.user.firstName ?
-          vm.user.firstName : vm.user.displayName;
+          vm.user.firstName.toLowerCase() : vm.user.displayName.toLowerCase();
```
davidmsibley: What's this about?

thevoiceofzeke: It's necessary to change the case of the username in the menu. It has to be programmatically changed to lowercase because the CSS `text-transform: capitalize` has no effect on capital letters. I figured using this with CSS is less overkill than making a custom filter.

PR #509, opened 2017-08-29
fix(widget): Use widget pref to determine target in option-link

davidmsibley:
Fixes option-link widget so it will use `widget.target` rather than hard coded `_self`

----

### DIFF
--- a/uw-frame-components/portal/widgets/partials/type__option-link.html
+++ b/uw-frame-components/portal/widgets/partials/type__option-link.html
@@ -2,7 +2,7 @@
 
   <!-- OPTION-LINK ICON -->
   <div class="option-link-icon">
-    <a href="{{ widget.selectedUrl }}" target="_self" rel="noopener noreferrer">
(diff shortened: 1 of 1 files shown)


uw-frame-components/portal/widgets/partials/type__option-link.html
```diff
@@ -2,7 +2,7 @@
 
   <!-- OPTION-LINK ICON -->
   <div class="option-link-icon">
-    <a href="{{ widget.selectedUrl }}" target="_self" rel="noopener noreferrer">
+    <a href="{{ widget.selectedUrl }}" target="{{ widget.target ? widget.target : '_self' }}" rel="noopener noreferrer">
```
apetro: I'm starting to be sold on the idea of auto-determining target depending upon the value of the `href`. As in, if it starts with `/`, it's `_self`; if it starts with `http`, it's `_blank` with the `rel="noopener noreferrer"`.

davidmsibley:
After some exploration, defining the target at the widget level isn't the most accurate way to go about it.

PR #519, opened 2017-09-11
feat(widget): use smart target on option-link widget

davidmsibley:
Using logic to set target rather than pulling from widget properties. It happens to be the dumbest logic, but still logic. https://tools.ietf.org/html/rfc3986#section-3

----

### DIFF
--- a/uw-frame-components/portal/misc/filters.js
+++ b/uw-frame-components/portal/misc/filters.js
@@ -65,6 +65,15 @@ define(['angular'], function(angular) {
         return filtered;
       };
     })
+    .filter('urlToTarget', function() {
+      return function(url) {
+        var result = '_self';
+        if (url && -1 < url.indexOf('//')) {
+          result = '_blank';
+        }
+        return result;
+      };
+    })
 
     /* WARNING: THIS FILTER IS DANGEROUS.
        You should only filter to trusted status HTML that you are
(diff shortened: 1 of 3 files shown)


uw-frame-components/portal/misc/filters.js
```diff
@@ -65,6 +65,15 @@ define(['angular'], function(angular) {
         return filtered;
       };
     })
+    .filter('urlToTarget', function() {
+      return function(url) {
+        var result = '_self';
+        if (url && -1 < url.indexOf('//')) {
```
ChristianMurphy: :question: can this ever be true?
[`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) will return -1 if not found and a number >= 0 if value is found. It will never be less than -1 AFAIK.

davidmsibley: I think so...
not found: -1 < -1 (false)
found: -1 < 0 or above (true)

ChristianMurphy: yep, my bad I misread that.
:art: avoid [Yoda Conditions](https://en.wikipedia.org/wiki/Yoda_conditions)

davidmsibley: Please express your concerns and reservations about code style and design choices, but I'd appreciate something more than an "avoid ______" statement and a link to Wikipedia.

ChristianMurphy: Okay, I usually use the 🎨 emoji has shorthand to say this it is a code style preference.
If you prefer long hand

> It is my personal preference that we should avoid [Yoda Conditions](https://en.wikipedia.org/wiki/Yoda_conditions) in the code style

PR #521, opened 2017-09-12
Notifications may be non-dismissible

thevoiceofzeke:
[MUMUP-3021](https://jira.doit.wisc.edu/jira/browse/MUMUP-3021): "As a user, I would like MyUW notifications to sometimes be non-dismissable, so that I am reminded more often and can't accidentally dismiss a notification that is attempting to notify me of a critical action I need to take."

**In this PR:**
- Add support for `notDismissible` attribute in messages data model
- Tweak notifications/announcements menu UI
    - Priority flag less disruptive of UI, but still noticeable
    - Announcement titles look like notification titles
    - Description text slightly lighter color
    - Dismiss button centered in announcement menu items
- If addToHome action has already been taken, disable button
- Update messaging configuration doc

#### notDismissible notification
screen shot 2017-09-12 at 2 21 32 pm
#### High-priority UI tweak
screen shot 2017-09-12 at 2 12 39 pm
#### Announcement item UI tweak
screen shot 2017-09-12 at 2 22 10 pm

----

### DIFF
--- a/docs/messaging-implementation.md
+++ b/docs/messaging-implementation.md
@@ -22,6 +22,7 @@ point uw-frame to your desired feed.
             "expireDate": "2017-08-02",
             "featureImageUrl": null,
             "priority": "high",
+            "isSticky": true,
             "audienceFilter": {
                 "groups": ["Users - Service Activation Required"],
                 "dataUrl": "/restProxyURL/unactivatedServices",
@@ -54,6 +55,7 @@ point uw-frame to your desired feed.
 - **expireDate**: *(optional)* Accepts a simple ISO date, including time (as pictured). This is used to stop displaying a message at a certain day/time.
 - **featureImageUrl**: *(optional)* Used by popup announcements and announcements on the Features page.
 - **priority**: Set to "high" if you want the message to be displayed with higher visibility (i.e. As a priority notification or popup announcement, respectively).
+- **isSticky**: *(experimental)* Set to true if you want to disallow users from dismissing a notification. This should only be used for truly critical messages. 
 - **audienceFilter**: A group of attributes related to filtering messages based on a user's group or whether the user has pertinent data at a given URL.
   - **groups**: An attribute to optionally show messages only to specific groups (i.e. uPortal groups). If null or empty array, the message will be shown to everyone. Contact your portal development team for more information about group filtering.
   - **dataUrl**: *(optional)* The message will retrieve data from the dataUrl. If data exists, it will show this message to the user. Only supports JSON.
--- a/docs/messaging-implementation.md
+++ b/docs/messaging-implementation.md
@@ -22,7 +22,7 @@ point uw-frame to your desired feed.
             "expireDate": "2017-08-02",
             "featureImageUrl": null,
             "priority": "high",
-            "isSticky": true,
+            "dismissible": false,
             "audienceFilter": {
                 "groups": ["Users - Service Activation Required"],
                 "dataUrl": "/restProxyURL/unactivatedServices",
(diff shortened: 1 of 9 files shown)


uw-frame-components/portal/messages/partials/notifications-bell.html
```diff
@@ -54,13 +54,13 @@
...
           <div class="menu-item-content" layout="row" layout-align="space-between center">
             <div flex="80" layout="column" layout-align="center start">
-              <!-- REGULAR MESSAGE -->
-              <span ng-if="!notification.priority">{{ notification.title }}</span>
               <!-- PRIORITY MESSAGE -->
-              <div layout="row" layout-align="start center" ng-if="notification.priority">
-                <md-icon aria-label="important:" flex="10">priority_high</md-icon>
-                <span flex="90"> {{ notification.title }}</span>
+              <div class="menu-item__high-priority" ng-if="notification.priority">
+                <md-icon class="md-warn" aria-label="important:">priority_high</md-icon>
```
apetro: Given the `<span>High priority</span>`, maybe this `md-icon` can be considered decorative and `aria-hidden`? Or `aria-labelledby` the span?

uw-frame-components/portal/messages/partials/notifications-bell.html
```diff
@@ -54,13 +54,13 @@
           <!-- NOTIFICATION CONTENT -->
           <div class="menu-item-content" layout="row" layout-align="space-between center">
             <div flex="80" layout="column" layout-align="center start">
-              <!-- REGULAR MESSAGE -->
-              <span ng-if="!notification.priority">{{ notification.title }}</span>
               <!-- PRIORITY MESSAGE -->
-              <div layout="row" layout-align="start center" ng-if="notification.priority">
-                <md-icon aria-label="important:" flex="10">priority_high</md-icon>
-                <span flex="90"> {{ notification.title }}</span>
+              <div class="menu-item__high-priority" ng-if="notification.priority">
```
apetro: Maybe should `ng-if` on `notification.priority` having the actual value "high", so that if we would in the future invent other priority values this high priority treatment wouldn't trigger on them.

uw-frame-components/staticFeeds/sample-messages.json
```diff
@@ -86,6 +91,7 @@
       "expireDate": "2017-08-15",
       "featureImageUrl": "img/the-cow.png",
       "priority": "high",
+      "notDismissible": false,
```
apetro: Let's not with the negative in the name of the property. Maybe `"userDismissible": true,` ?

thevoiceofzeke: My thinking here was that the default case should be that dismissible is true, since the vast majority of notifications are not important enough to linger. The attribute should be optional and should only be set when something's important enough to be non-dismissible -- otherwise folks shouldn't have to bother with it. I agree that the negative property is a bad look. What if we used something to reflect non-dismissibility, like "permanent" or "sticky"?

docs/messaging-implementation.md
```diff
@@ -54,6 +55,7 @@ point uw-frame to your desired feed.
 - **expireDate**: *(optional)* Accepts a simple ISO date, including time (as pictured). This is used to stop displaying a message at a certain day/time.
 - **featureImageUrl**: *(optional)* Used by popup announcements and announcements on the Features page.
 - **priority**: Set to "high" if you want the message to be displayed with higher visibility (i.e. As a priority notification or popup announcement, respectively).
+- **isSticky**: *(experimental)* Set to true if you want to disallow users from dismissing a notification. This should only be used for truly critical messages.
```
apetro: Consider documenting default behavior.  Maybe "By default users can dismiss notifications. Set to true to disallow user dismissing for this notification."

"This should only be used for truly critical messages. ": I'm not sure we know that this is the guidance yet. Stickiness may be less about criticality and more about does the notification dismiss itself when the underlying business process is fulfilled. If we can tell that you no longer have a library fine preventing your graduating, we should yank the notification for you when you pay the fine. Maybe how well we can automate this for the user drives how much dismissing the notification work we ought to expose to the user.

(Notifications are like a to do list where the items check themselves off when you do them?)

We might get to guidance that whether user dismissing should be disabled depends only on the criticality of the message. I think the point of the prospective user research is to figure this out.

uw-frame-components/portal/messages/partials/notifications-list-item.html
```diff
@@ -8,7 +8,7 @@
    layout-align="start center"
    layout-xs="column"
    layout-align-xs="center start">
-  <span><md-icon ng-if="notification.priority">priority_high</md-icon> {{ notification.title }}</span>
+  <span><md-icon ng-if="notification.priority == 'high'">priority_high</md-icon> {{ notification.title }}</span>
```
apetro: How do non-visual users understand that the notification is a priority notification?

PR #531, opened 2017-09-21
NOTICE and pom refactor

davidmsibley:
Moving the pom down to the root level of the project, which is where the jasig-parent expects it to be. Also, jasig-parent is picky about what the NOTICE says.

----


apetro:
:+0: `jasig-parent` is outdated about what should be in the `NOTICE` file. Didn't keep up with evolving Apache Software Foundation practice on this. Which is, I suppose, a problem with Jasig/Apereo having our own tools for this stuff. And even having our own fork of Apache policy rather than just adopting Apache policy by reference. Lots of opportunity for improvement.

So. Not gonna stand in the way of progress. Fine, we merge this, cut a release, get the repo moved, all good. Someday we stop using Maven, and then we can stop being subject to the outdated `NOTICE` checking plugin, and then we can have a `NOTICE` file that doesn't include [acknowledgements it doesn't need](http://www.apache.org/dev/licensing-howto.html#mod-notice).

NOTICE
```diff
@@ -1,6 +1,20 @@
-Apereo uPortal-application-framework aka uw-frame
+Licensed to Apereo under one or more contributor license
+agreements. See the NOTICE file distributed with this work
```
apetro: This *is* the `NOTICE` file distributed with this work. :)

docs/skinning.md
```diff
@@ -117,7 +117,7 @@ For a more in-depth discussion of theming, see <https://github.com/UW-Madison-Do
 
 7.  In your tools/uw-frame-static/build.js file, add your theme's name to the array of themes.
 
-8.  From a command line, navigate to your uw-frame-java directory, and run
+8.  run
     ```sass
```
ChristianMurphy: :question: should this be flagged as a shell script?

PR #558, opened 2017-10-03
fix(theme): using uw-system as default theme

davidmsibley:
We had an issue with my.wisconsin.edu defaulting to the uPortal theme rather than uw-system. On top of that, search/marketplace is broken with the uPortal theme ( #559 ). We're not quite operationally ready for uPortal to be the default theme yet, so this changes it back to uw-system.

----

### DIFF
--- a/components/js/frame-config.js
+++ b/components/js/frame-config.js
@@ -24,12 +24,12 @@ define(['angular'], function(angular) {
          * THOU SHALT INCREMENT THIS VERSION NUMBER
          * IF THOU CHANGEST ANY OF THE THEMES BELOW
          */
-        'themeVersion': 9,
+        'themeVersion': 10,
         'themes': [
           {
             'name': 'default',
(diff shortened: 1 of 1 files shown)


davidmsibley:
yup, just wanted to put a PR in while we're thinking about it to reflect what we did locally.

An even further-looking solution may be that we use this as an opportunity to de-UW-ify the underlying codebase, and we pull these themes out of the codebase and into our institution-specific overlay.

apetro:
I agree there are advantages in de-UW-ifying the open source product. Let's maybe think about how to do that in a way that still keeps more real-world examples available to all would-be `uPortal-home` adopters. Keeping interesting customizations private is an aspect of how we got all forked in our underlying `uPortal`. Maybe a layer of UW-ification could also be in a public repo or in a public library? Or maybe not worth the complexity -- the UW-ifications are way different in kind and scale than some of our uPortal customizations.

PR #560, opened 2017-10-03
fix: Redesign mobile search

thevoiceofzeke:
[MUMUP-3099](https://jira.doit.wisc.edu/jira/browse/MUMUP-3099): "As a mobile user of MyUW, I'd like the search button to do something, so that I'm not confused when I click the button and nothing happens."

**In this PR**:
- Refactored behavior of search directive
- Redesigned appearance of mobile search field to take up less space

### Screenshot
screen shot 2017-10-03 at 2 28 44 pm

### Demo
search-fix

*To do: Animate open/close of mobile search*

----

Contributor License Agreement adherence:

- [x] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

[Apereo CLA roster]: http://licensing.apereo.org/completed-clas
[Contributor License Agreements]: https://www.apereo.org/licensing/agreements

PR #561, opened 2017-10-03
Redesign top bar menu and add side navigation

thevoiceofzeke:
This is the result of experimenting with input/request from other groups on campus. Marking it WIP because I want some eyes on it before I spend significant time testing/documenting.
 
**In this PR**:
- Top bar hamburger menu always visible
- Hamburger menu button triggers side navigation
- Side nav closes when clicking outside or when scrolling
- Side navigation is configurable via `APP_OPTIONS.appMenuItems` 
- Side navigation subheader (above the options) uses the configurable `NAMES.title` (aka the app title) to differentiate it from `portal.theme.title` (aka MyUW/uPortal)
- On mobile, side navigation overlays top bar and includes username menu, notification button, and features button (if applicable)
- Replace unused directive with `mainMenu`
- If no custom menu items are configured, side nav only displays on mobile (with only the basic uportal-app-framework content and footer URLs)
- Trigger priority notifications from PortalMainController instead of PortalHeaderController, since more things than the header need to know about this (and that controller hierarchy was making things difficult)
- Includes the changes from #560
- No more crest image. I expect this to be divisive, but I feel strongly that the crest image provides very little benefit while using valuable space and increasing the project's payload.
- Document how to use new feature

### Demos (large/small)
sidenav-large
sidenav-small

### Screenshots
screen shot 2017-10-03 at 3 14 58 pm
screen shot 2017-10-03 at 3 17 33 pm

screen shot 2017-10-11 at 11 51 28 am

screen shot 2017-10-03 at 3 17 59 pm
screen shot 2017-10-03 at 3 18 09 pm
screen shot 2017-10-03 at 3 18 18 pm

#### Without high priority notifications or unseen announcements

screen shot 2017-10-03 at 3 19 35 pm

----


apetro:
Not gonna lie: I miss the crest. It seems important to branding a portal as institutional, to reassuring that this is "yours", that it's your relationship with the campus. At Madison, it's "MyUW". The "UW" part of that is in part signaled by the crest.

thevoiceofzeke:
@apetro I wouldn't be against bringing the crest back, but I think finding an appropriate place to put it is a challenge. The current spot is too much of a blocker of functionality (and we already don't show it on mobile for real estate reasons), and I feel that the benefits of an improved navigation scheme (especially as it concerns developers of other frame apps) outweigh the value of having the crest there. 

One immediate change I can include in this PR is showing the crest at the top of the side navigation menu on mobile, so it'll actually have more presence on mobile that it currently does in production.

Would you be able to get on board with considering it a temporary removal, pending some design brainstorming? I agree that it has important branding value and strengthens the app's sense of place, so I'd like to find a way to include it that doesn't block functionality.

apetro:
I'm already on board, @thevoiceofzeke . I do predict that UX design input, MyUW service team feedback, is going to bring that crest back sooner or later. :) But I'm more than fine with letting the experiment play out, prioritizing developing usable navigation and revisiting branding thereafter. Either the crest comes back in the course of that or it doesn't, and either way will be ok.

cknuth:
I'm excited for the side navigation menu! That's something we have struggled with incorporating in STAR and would love to take advantage of.

Rather than a list of links for the navigation, it would be great if we could have a more fine grained control as to what shows in that side menu. For instance, we want to remember search params for some pages in our app which isn't possible with a static link. Furthermore, it would be great to group the links into sections and hide/show those sections based a user's authorities.

If the side nav had an option for a template then we could take matters into our own hands. We've found that we'v been blocked by the limitations of the what we could change in the frame, but having flexibility in these areas would make me more likely to choose the app-framework for IA apps in the future.

components/js/override.js
```diff
@@ -22,6 +22,14 @@ define(['angular'], function(angular) {
     .constant('OVERRIDE', {
       'APP_OPTIONS': {
         'optionsTemplateURL': 'portal/misc/partials/example-options.html',
+        'appMenuTemplateURL': 'portal/misc/partials/example-menu.html',
```
apetro: What if the API were a URL that returned JSON representing the menu? Once it's on the other end of a URL, it can be arbitrarily implemented, whether as a static file or as a dynamic back end that considers user identity.

I think I'd prefer to define a semantic JSON API for menu and have the `uportal-app-framework` render that rather than escape to an arbitrary template HTML, so that we can semantically version the API. We can then engineer the framework to continue to support the JSON across versions of the framework, regardless of the evolution of the implementation of the menu.

components/portal/main/partials/header.html
```diff
@@ -19,39 +19,14 @@
...
-          <!-- Log out link -->
-          <md-menu-item>
-            <md-button class="md-default" ng-href="{{MISC_URLS.logoutURL}}" target="_self" ng-click="pushGAEvent('Mobile Menu', 'Click Link', 'Log out');">
-              <md-icon>exit_to_app</md-icon> Log out
-            </md-button>
-          </md-menu-item>
-        </md-menu-content>
-      </md-menu>
+      <!-- MAIN MENU TOGGLE -->
+      <md-button class="menu-toggle" ng-controller="MainMenuController as vm" ng-click="vm.showMainMenu()" ng-class="{ 'hide-gt-xs' : vm.hideMainMenu }">
```
apetro: Does this button need some kind of ARIA treatment?

components/js/override.js
thevoiceofzeke: I agree with you. This quick and dirty option is meant to meet two of those requirements that have already come up in conversation:

- Show/hide items depending upon user attributes or group memberships
- Menu items that call JavaScript rather than just hyperlink a URL

I'm having a hard time envisioning how all those pieces would come together, but I would like to establish a better way of doing this.

thevoiceofzeke:
@vertein I've been wondering about embracing UW-Madison's fonts for a while (even if extremely sparingly). Seems like the top bar would be a good place to try it.

apetro:
This may have introduced #582 .

PR #569, opened 2017-10-10
feat: adding filter to calculate rel on anchors

davidmsibley:
Adding a filter to calculate whether `noopener noreferrer` should be added to an anchor based on the target value.

----

### DIFF
--- a/components/portal/misc/filters.js
+++ b/components/portal/misc/filters.js
@@ -98,7 +98,8 @@ define(['angular'], function(angular) {
     .filter('targetToRel', function() {
       return function(target) {
         var result = '';
-        if (target === TARGET_BLANK) {
+        if (target &&
+            target.trim().toUpperCase() === TARGET_BLANK.toUpperCase()) {
           result = REL_BLANK_FIX;
         }
         return result;
--- a/components/portal/misc/filters.js
+++ b/components/portal/misc/filters.js
@@ -20,6 +20,9 @@
 
 define(['angular'], function(angular) {
     var DEFAULT_TRUNCATE_LENGTH = 160;
+    var TARGET_SELF = '_self';
+    var TARGET_BLANK = '_blank';
+    var REL_BLANK_FIX = 'noopener noreferrer';
     return angular.module('portal.misc.filters', [])
 
     .filter('truncate', function() {
@@ -85,9 +88,18 @@ define(['angular'], function(angular) {
     })
     .filter('urlToTarget', function() {
       return function(url) {
-        var result = '_self';
+        var result = TARGET_SELF;
         if (url && url.indexOf('//') > -1) {
-          result = '_blank';
+          result = TARGET_BLANK;
+        }
+        return result;
+      };
+    })
+    .filter('targetToRel', function() {
+      return function(target) {
+        var result = '';
+        if (target === TARGET_BLANK) {
+          result = REL_BLANK_FIX;
         }
         return result;
       };


davidmsibley:
is used in https://github.com/uPortal-Project/uportal-home/pull/725

components/portal/misc/filters.js
```diff
@@ -85,9 +88,18 @@ define(['angular'], function(angular) {
     })
     .filter('urlToTarget', function() {
       return function(url) {
-        var result = '_self';
+        var result = TARGET_SELF;
         if (url && url.indexOf('//') > -1) {
-          result = '_blank';
+          result = TARGET_BLANK;
```
ChristianMurphy: :art: could this be directly `return`ed, rather than stored to intermediate `result`?

components/portal/misc/filters.js
```diff
@@ -85,9 +88,18 @@ define(['angular'], function(angular) {
...
+        }
+        return result;
```
ChristianMurphy: :art: could this directly return `TARGET_SELF`?

components/portal/misc/filters.js
```diff
@@ -85,9 +88,18 @@ define(['angular'], function(angular) {
...
+      };
+    })
+    .filter('targetToRel', function() {
+      return function(target) {
+        var result = '';
+        if (target === TARGET_BLANK) {
+          result = REL_BLANK_FIX;
```
ChristianMurphy: :art: could this be directly `return`ed, rather than stored to intermediate `result`?

davidmsibley:
@ChristianMurphy I'm of the opposite mind, that there should only ever be one return statement per function

components/portal/misc/filters.js
```diff
@@ -85,9 +88,18 @@ define(['angular'], function(angular) {
...
-          result = '_blank';
+          result = TARGET_BLANK;
+        }
+        return result;
+      };
+    })
+    .filter('targetToRel', function() {
+      return function(target) {
+        var result = '';
+        if (target === TARGET_BLANK) {
```
apetro: In the interest of having a generically usable `targetToRel` filter, should this be equals ignoring case? Possibly after trimming whitespace from `target`?  Then one could pipe arbitrary `<a` tag `target` attributes in the world through `targetToRel` to get the appropriate `rel` for that anchor?

vertein: I see trimming being useful, not as much of ignoring case.  But if you're going to do one, it's not hard to do both.

davidmsibley: we certainly could, but actually we probably shouldn't calculate off of target at all. Since the concern comes from external sites being hijacked, a more foolproof option would be to calculate the rel from the url, like we do in the filter right above this. Or we could straight up just add the rel to every link we create, provided internal pages don't require the referrer

apetro: Hmm. But it's the `target`, not the URL, that triggers the vulnerability, isn't it? A `target` of `_blank` or `_BLANK` or `_BlAnK` or, browsers being what they are, probably `_blank   ` is what sets up the opportunity for the newly opened tab to mess with the prior no-longer-focus-of-attention tab location, a messing-with that's blocked by the `rel` incantation.

davidmsibley: ahhh, ok. I have a superficial understanding of the vuln. Yup, I'll clean up the target matching

ChristianMurphy:
> I'm of the opposite mind, that there should only ever be one return statement per function
> I like the intermediate step of saving the value before returning in this case as it helps me with readability.

If the code was multiple-entry multiple-exit (GOTO/LABEL/JUMP) or if there were side-effects in the code, I would completely agree.
These are [pure functions](https://en.wikipedia.org/wiki/Pure_function), when that is the case, I tend to lean toward the functional programming principle that variables are to be avoided.
But I digress. :wink: 

Functionally the change looks good. :+1:

PR #575, opened 2017-10-12
fix: Fix app hang when promise returns undefined

thevoiceofzeke:
After pointing star app at new messages feed, the service fails to get messages (a bigger problem than this PR addresses). When it fails, the angular controls hangs because the promise returns `undefined`. This PR just makes sure the block doesn't continue if the promise result is undefined.

----

### DIFF
--- a/components/portal/messages/controllers.js
+++ b/components/portal/messages/controllers.js
@@ -42,11 +42,13 @@ define(['angular'], function(angular) {
         var getMessages = function() {
           messagesService.getAllMessages()
             .then(function(result) {
-              // Ensure messages exist and check for group filtering
-              if (result.messages && result.messages.length > 0) {
-                allMessages = result.messages;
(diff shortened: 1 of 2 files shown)


components/portal/messages/controllers.js
```diff
@@ -42,11 +42,13 @@ define(['angular'], function(angular) {
         var getMessages = function() {
           messagesService.getAllMessages()
             .then(function(result) {
-              // Ensure messages exist and check for group filtering
-              if (result.messages && result.messages.length > 0) {
-                allMessages = result.messages;
+              if (result) {
```
ChristianMurphy: :question: should this be handled at the `messagesService` layer?
replacing `underfined` with `[]`?

:question: could this be labeled with `TODO` or `FIXME` as a reminder that it is a short term fix?

PR #576, opened 2017-10-13
fix: Include /web check in notifications url eval

thevoiceofzeke:
**In this PR**:
- First checks if configured notifications url points to portal base path (`/web`), then evaluates `$location.path()`
- Fixed a bug where incorrect type check was causing notifications feed to always be empty

This is a little hacky, but the idea is if a frame app wants to point to the portal-wide notifications feed, it has to point to /web. Since /web is the portal base path (and not included in any $location getters), evaluating the path against the configured notificationsPageURL will always result in `false`. We have to check for /web, and it it's not part of the configured path, we're free to fall back on whatever other path was configured. 

----

### DIFF
--- a/components/portal/messages/controllers.js
+++ b/components/portal/messages/controllers.js
@@ -125,10 +125,10 @@ define(['angular'], function(angular) {
       },
     ])
 
-    .controller('NotificationsController', ['$q', '$log', '$scope', '$browser',
+    .controller('NotificationsController', ['$q', '$log', '$scope',
       '$rootScope', '$location', '$localStorage', '$filter', 'MESSAGES',
       'SERVICE_LOC', 'miscService', 'messagesService',
-      function($q, $log, $scope, $browser, $rootScope, $location, $localStorage,
+      function($q, $log, $scope, $rootScope, $location, $localStorage,
                $filter, MESSAGES, SERVICE_LOC, miscService, messagesService) {
         // //////////////////
         // Local variables //
--- a/components/portal/messages/controllers.js
+++ b/components/portal/messages/controllers.js
@@ -43,8 +43,8 @@ define(['angular'], function(angular) {
           messagesService.getAllMessages()
             .then(function(result) {
               // Ensure messages exist and check for group filtering
-              if (result.messages && result.messages.length > 0) {
-                allMessages = result.messages;
+              if (result.length > 0) {
+                allMessages = result;
               }
               filterMessages();
               return allMessages;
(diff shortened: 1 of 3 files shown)


components/portal/messages/controllers.js
```diff
@@ -175,6 +175,20 @@ define(['angular'], function(angular) {
         // Local methods //
         // ////////////////
         /**
+         * Check if user is viewing notifications page
+         */
+        var isNotificationsPage = function() {
+          if (MESSAGES.notificationsPageURL.indexOf('/web') != -1) {
```
ChristianMurphy: :art: prefer strict equality check.

:thought_balloon: I have a dream that one day `Array.includes` will be supported by the browser matrix.

components/portal/messages/controllers.js
```diff
@@ -128,8 +128,8 @@ define(['angular'], function(angular) {
     .controller('NotificationsController', ['$q', '$log', '$scope',
       '$rootScope', '$location', '$localStorage', '$filter', 'MESSAGES',
       'SERVICE_LOC', 'miscService', 'messagesService',
-      function($q, $log, $scope, $rootScope, $location, $localStorage, $filter,
-               MESSAGES, SERVICE_LOC, miscService, messagesService) {
+      function($q, $log, $scope, $rootScope, $location, $localStorage,
+               $filter, MESSAGES, SERVICE_LOC, miscService, messagesService) {
```
ChristianMurphy: :star: nice catch

davidmsibley:
Instead of hardcoding `/web` in there, can we find a way to use `$window.location.path` rather than `$location.path`? `$window.location` includes the base href

thevoiceofzeke:
@davidmsibley I was looking for `$window.location.pathname`!. Ugh. It is extremely aggravating that the $window service and $location service use the same API but differ ever so slightly in what they deliver. On uportal-home's landing page, for example, `$location.path() === /expanded` and `$window.location.pathname === /web/expanded`.

WHY

davidmsibley:
hah! Glad you caught that

ChristianMurphy:
> Ugh. It is extremely aggravating that the `$window` service and `$location` service use the same API but differ ever so slightly in what they deliver
> ...
> WHY

`$location` was designed as an abstraction for routing within Angular.
Its methods are designed to work the same if the app is running on the root path of a web server or under a partial path.

`$window` is a direct wrapper of the `window` object, it is mostly designed as a way to allow communication with non-angular components, but can also be used to directly access globals like `location`.

The idiosyncrasy comes from Angular 1's sometimes helpful, sometimes frustrating attempt to support multiple programming styles.
A similar quirk would be [`factory`](https://docs.angularjs.org/guide/providers#factory-recipe), [`service`](https://docs.angularjs.org/guide/providers#service-recipe), and [`provider`](https://docs.angularjs.org/guide/providers#provider-recipe).

Issue #535, opened 2017-09-22
A plea to abolish commit hooks


davidmsibley:
I realize I've expressed being against commit hooks, but I don't think I've explained why.

Committing is always good. I never want to discourage people from committing. I realize there's value in influencing the quality of code, but I believe that investment should be made at the integration layer (CI and PR's). Version control should be about retaining history, so that we can learn how code came about. You can't do that if your version control stops you from storing your work until you've made it presentable.

In the same vein, I want to advocate using pull requests as learning opportunities or teachable moments, rather than customs checkpoints. No one's going to lose their job because they didn't put curlies around an if block. So, the Watchmen of this little world, we ain't. Instead, Committers get to be shepherds and explorers within the world of code. We get to learn from one another and improve our skills as we grow the value of the product. If we just enforce rules, it's difficult to learn why those rules are effective. So again, I'd encourage giving developers access to the tools that improve our codebase, and impressing upon them the reasons we value each tool.

Finally, imperfections within the code (in moderation) are often the best motivation to improve your craft. Refactoring is the most effective method of learning how and why a system works as it does. Teaching developers to spot potential problems encourages them to refactor early and often. Letting imperfect code through the integration process can drive developers to grow, while still keeping a handle on a product's technical debt.

I think the style guides and the linting have definitely given this product a lot of value, but I want to voice concern before we head too far down this path. Every project has a different set of circumstances and requirements. I'd like to make sure everyone who works on this project can navigate us towards the destination, rather than just being able to keep us on the road. I want to encourage committing early and often, using pull requests as learning opportunities, and giving developers the fire to improve. I believe this is the best way to build a community that values the same things, not one that conforms to the style du jour.

ChristianMurphy:
I hear you @davidmsibley. 👂 

I completely agree that uPortal-app-framework should strive to be as easy to contribute to and as welcoming as possible.  💯 
I still believe that Git Hooks help enable and empower the contributing process, rather than hurting it.
[-0.9](https://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions) for removing Git Hooks.

---

> Version control should be about retaining history, so that we can learn how code came about. You can't do that if your version control stops you from storing your work until you've made it presentable.

That is one way of looking at it.
From another perspective, for git history to have value, you need to be able to see what people *really* changed, tools like `git blame` can't tell the difference between style fixes, features, and bug fixes.
[...]

PR #580, opened 2017-10-15
style(js,json,less): use 'prettier' to handle code style

ChristianMurphy:
Rather than throwing an error during `precommit` hook.
`prettier` will reformat all the code to meet the styleguide standards automatically.

Check out this talk for more information on the 'why' to use prettier.
<a href="https://www.youtube.com/watch?v=hkfBvpEfWdA"></a>

part 1 of #535
part 2 is in https://github.com/uPortal-Project/uportal-app-framework/pull/581

----


apetro (approved):

davidmsibley (changes requested):
This looks like it's replacing single quotes to double quotes, breaking the google style guide?

ChristianMurphy:
@davidmsibley Currently I have `prettier` setup to use all of its defaults.
Those defaults are very similar, but subtly different than the Google Code style, we can adjust the settings to use single quotes.

ChristianMurphy:
@davidmsibley updated to use single quotes

davidmsibley:
I'm confused about a couple things in this. I gotta read it more, but I was referring to the single to double quote replacement that was happening in the js, not the less. Now the less is switched from double to single quotes.

Unrelated to whether or not we use prettier, I'm WAY more opposed to code _being changed_ on commit than stopping you from committing. That sounds like many, many unintended problems waiting to happen.

ChristianMurphy:
:laughing: The codeclimate failure is occurring because the normalized code style allows codeclimate to detect nine places with code similarity that it wasn't able to before.

vertein:
Does this automatically re-format to use Google's code style or does this use a separate style that we would have to match to Google's?

apetro:
> subtly different than the Google Code style

Okay, I hadn't understood that nuance. -1.  Code is either in Google Style or it's not. Automatically different, even subtly, is not Google Style. I'd only be open to automatic code styling if it's automatically styling to the adopted code style. :)

This also may be a tool that it'd be best to adopt on a less-core module or three before adopting it here in `uportal-application-framework`. I'm not sure I share @davidmsibley 's prediction of many unintended problems, but I certainly see how a reasonable person would fear that. I think we can be way more open to experimental friction and risk in modules that we're less likely to need to release frequently *for adoption by other teams*, which is what this module is.

ChristianMurphy:
> Code is either in Google Style or it's not. Automatically different, even subtly, is not Google Style.

> Does this automatically re-format to use Google's code style or does this use a separate style that we would have to match to Google's?

The subtle differences are trailing commas and quotes.
Both can be configured to match Google Code Style. :white_check_mark: 

> I'd only be open to automatic code styling if it's automatically styling to the adopted code style.

Or we could make the automated style the adopted style. :wink: 

> That sounds like many, many unintended problems waiting to happen.

I understand the concern, in the past formatters could be flaky. :fearful: 

Prettier and ESLint are based of [ESTree](https://github.com/estree/estree) an [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) standard [developed by Mozilla](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API) for accurately representing and safely manipulating code structure. :+1: 
Using AST for code manipulation resolves the invalid parsing and serialization errors common in regular expression based parsers (which did have major issues with producing invalid code). :white_check_mark: 

Furthermore, Prettier is in production use today on some major projects: [React](https://github.com/facebook/react/blob/101934646f79660430a5e76487d6e5350e68a7eb/package.json#L103), [Jest](https://github.com/facebook/jest/blob/027622436da65957e8f2ff31ac2e222855a00421/package.json#L58), [Yarn](https://github.com/yarnpkg/yarn/blob/b97c797983babd03c68bc36f1ee98a9a6dff9246/package.json#L88), [and others](https://prettier.io/en/users/).
The software is pretty well battle tested. :crossed_swords: 

> This also may be a tool that it'd be best to adopt on a less-core module or three before adopting it here in `uportal-application-framework`

:man_shrugging: With the clean breakup of modules, it is arguably questionable if there is a "core" and what isn't core. Until recently I thought [`uPortal-home` was "core"](https://github.com/uPortal-Project/uportal-home/pull/704#discussion_r141414623).

> I think we can be way more open to experimental friction and risk in modules that we're less likely to need to release frequently for adoption by other teams, which is what this module is.

Since all the modules are closely interrelated, what is module is sufficiently "low risk" in your mind?

apetro:
+ `uPortal-home` is lower risk than `uPortal-app-framework`, in that only the latter has downstream projects depending upon it.
+ [`widget-creator`](https://github.com/UW-Madison-DoIT/widget-creator) is much lower risk, in that its audience is developers.

ChristianMurphy:
Closing this, til decision is made over in https://github.com/UW-Madison-DoIT/widget-creator/pull/25

PR #583, opened 2017-10-17
Fix spacing when app navigation isn't configured

thevoiceofzeke:
**In this PR**:
- Added some CSS to maintain margins with or without side nav menu button
- Removed a little unused CSS
- Autohide menu tooltip so it doesn't linger on mobile

----

### DIFF
--- a/components/css/buckyless/header.less
+++ b/components/css/buckyless/header.less
@@ -44,19 +44,6 @@ portal-header {
       max-height: 56px;
       overflow-y: hidden;
 
-      .title-link {
-        h1 {
-          @media (min-width: @xs) {
-            margin-left: 16px;
-          }
-        }
-
-        img {
-          height: 56px;
-          margin-right: 8px;
-        }
-      }
-
       .md-menu {
         > .md-button {
           min-width: 48px;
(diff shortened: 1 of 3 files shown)


components/css/buckyless/header.less
```diff
@@ -67,6 +54,20 @@ portal-header {
       .menu-toggle.md-button {
         min-width: 48px;
         margin: 0;
+
+        + .title-link h1 {
```
ChristianMurphy: :question: would a [general sibling selector](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_selectors) be more robust?

components/css/buckyless/header.less
```diff
@@ -67,6 +54,20 @@ portal-header {
...
+          @media (min-width: @xs) {
+            margin-left: 0;
+          }
+        }
+
+        &.hide-gt-xs {
+          + .title-link h1 {
```
ChristianMurphy: :question: would a [general sibling selector](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_selectors) be more robust?

vertein:
@thevoiceofzeke, I'm going to cut a release soon.  This seems good to get in quick via merging?

PR #585, opened 2017-10-20
fix: Fix search widget type

vertein:
The action url wasn't appearing because the variable name was mistyped.

Compare with [variable name in controller](https://github.com/uPortal-Project/uportal-app-framework/blob/5e78410d1d816e0d030eb366c8bed10f1830df01/components/portal/widgets/controllers.js#L190)

----

### DIFF
--- a/components/portal/widgets/partials/type__search-with-links.html
+++ b/components/portal/widgets/partials/type__search-with-links.html
@@ -19,7 +19,7 @@
 
 -->
 <div class="widget-body">
-  <form action='{{ secureUrl }}' target='{{ config.actionTarget }}'>
+  <form action='{{ secureURL }}' target='{{ config.actionTarget }}'>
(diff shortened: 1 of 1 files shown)


components/portal/widgets/partials/type__search-with-links.html
```diff
@@ -19,7 +19,7 @@
 
 -->
 <div class="widget-body">
-  <form action='{{ secureUrl }}' target='{{ config.actionTarget }}'>
+  <form action='{{ secureURL }}' target='{{ config.actionTarget }}'>
```
apetro: I think `secureUrl` better upholds [Google Style](https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined). 

> Now lowercase everything (including acronyms), then uppercase only the first character

PR #588, opened 2017-10-26
Add on-page side navigation feature

thevoiceofzeke:
This is a feature that has been requested by a couple development teams working on uportal-app-framework applications. 

**In this PR:**
- Adjusted HTML hierarchy to allow main menu to sit alongside content
- Side navigation menu now part of the `<frame-page>` directive (thus is dependent on frame-page to work)
- Added APP_OPTIONS configuration option `enablePushContentMenu`. Only affects medium and large screen sizes. When set to true:
    - Side navigation is open on initial page load
    - Menu has transparent background (looks like part of the page)
    - Page content is pushed to the right
    - Page content slides back to be centered on screen when menu is closed
- Documented how to use new feature
- Adjusted styling for DOM parent elements to rely less on explicit `height` (favoring flex positioning to fill space)
- Misc. formatting fixes in markdown files (why why _why_ are there spaces before the colons?)

**Potential problem areas (UX):**
- Should menu button have "active" state indication?
- Should frame-page be built-in (i.e. not optional)? Making it part of the app structure (before custom content is transcluded) would mean the app-header and side navigation are always present and would enable us to greatly reduce the number of parent elements in HTML.

### Screenshot
screen shot 2017-10-26 at 10 21 09 am

### Demo 
sidenav-on-page-lg
sidenav-on-page-sm

#### Disclaimer

This implementation is a bit messy and will likely cause downstream breaking changes that are difficult to anticipate. Most of the hurdles here have to do with working around the existing app architecture (which is something framework devs have complained about in the past). Working on it has brought to light some convincing arguments for further simplifying (i.e. shrinking) the framework's footprint. With the addition of the side navigation, there are some steps we can safely take (probably) that will enable us to create more "portal-wide" features, improve the ones we have, and significantly clean up and reduce the HTML, JS, and CSS payload of this project. I've backlogged the following stories for discussing/grooming these things:
- [MUMMNG-3978](https://jira.doit.wisc.edu/jira/browse/MUMMNG-3978): Remove the app-header 
- [MUMMNG-3979](https://jira.doit.wisc.edu/jira/browse/MUMMNG-3979): Remove the built-in footer

----


mrdahman:
Usability testing of the design with four users showed us that users know how to interact with the side navigation, and other apps are interested in using the feature. Thumbs up.

docs/configurable-menu.md
```diff
@@ -4,18 +4,22 @@ When developing an app with uPortal-app-framework, you can use the configurable
...
 
-There are two levels of complexity you can use when configuring your menu.
+There are two levels of complexity you can use when configuring the contents of your menu.
+
+1. [Simple](#simple-configuration): Menu options that have a text label, a url, and an optional material icon (i.e. basic links)
+2. [Custom template](#custom-menu-template): Varied menu options, some of which may be tied to an angular controller and affect functionality of the app (e.g. a theme-switching toggle)
 
-1. Simple: Menu options that have a text label, a url, and an optional material icon (i.e. basic links)
-2. Complex: Varied menu options, some of which may be tied to an angular controller and affect functionality of the app (e.g. a theme-switching toggle)
+You can also try the _**experimental**_ [push-content feature](#push-content-navigation) if you want the side navigation to be open upon page load and appear as part of the page content. The side nav will push content when it is opened and give up the space when closed.
```
apetro: "Experimental" is interesting. If it's just "experimental", then maybe this doesn't have to be a major version bump? If we suspect we're going to break downstream apps tho, I'd favor the major version bump to signal this.

thevoiceofzeke: I mainly meant that as a "take caution using this" because it may be difficult to work with (due to the way it affect on-page content). I'm open to saying "caution" instead of "experimental" if you think that would be better. I do intend for this to be a permanent change.

prtreige:
I like it! Would it be possible to see click and active state when you select a navigation item for both the slide in and over navigation? In the slide in menu, I'd also like just a bit more vertical padding between the first item and the app name grey bar. Thanks, @thevoiceofzeke !

thevoiceofzeke:
@prtreige: There is currently no active state, in keeping with the implementation of angular material. I think there's room for debate that there _should_ be an active state, but that's perhaps better left for after a round of user testing. I'd say the same goes for vertical padding and other fine-tuning decisions. This PR is more about the bare minimum functionality. Thanks for your input!

docs/directives.md
```diff
@@ -1,12 +1,42 @@
 # AngularJS directives in uportal-app-framework
 
+## Frame page
+
+Frame page should be used for all pages in a uportal-app-framework application. It comes with much of the rest of the framework's skeleton built in, including the app-header and main-menu directives. You should use it as the outermost element for
```
ChristianMurphy: :art: `app-header` and `main-menu` directives should have some form of code highlight, it looks like `**` is being used elsewhere.

PR #589, opened 2017-10-26
feat: Recurring priority notifications

Doug-Reed:
Causes a notification to reoccur every time a user opens a new session, until the user falls out of the targeted group or the notification expires.

----

### DIFF
--- a/docs/messaging-implementation.md
+++ b/docs/messaging-implementation.md
@@ -22,7 +22,7 @@ point uportal-app-framework to your desired feed.
             "expireDate": "2017-08-02",
             "featureImageUrl": null,
             "priority": "high",
-            "recurrance": true,
+            "recurrence": true,
             "dismissible": false,
             "audienceFilter": {
                 "groups": ["Users - Service Activation Required"],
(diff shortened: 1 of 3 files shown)


docs/messaging-implementation.md
```diff
@@ -22,6 +22,7 @@ point uportal-app-framework to your desired feed.
             "expireDate": "2017-08-02",
             "featureImageUrl": null,
             "priority": "high",
+            "recurrance": true,
```
apetro: Spelling mismatch between here, the property in `services.js`, and the documentation below.

docs/messaging-implementation.md
```diff
@@ -22,6 +22,7 @@ point uportal-app-framework to your desired feed.
...
             "dismissible": false,
```
apetro: Consider making the `dismissible` property a String with several meaningful values ("false", "session", "true").  Recurrence feels more like a nuance of `dismissible` than a separate property? What would it mean for a notification to be non-dismissible but have recurrence set to `true`?

It may be possible to do this in a backwards-compatible way such that plain old `true` and `false` for `dismissible` continue to work, but even if it's not, that's probably ok because `dismissible` is called out as `experimental`. I don't think we have to semantically version experimental aspects of APIs.

docs/messaging-implementation.md
```diff
@@ -55,6 +56,7 @@ point uportal-app-framework to your desired feed.
 - **expireDate**: *(optional)* Accepts a simple ISO date, including time (as pictured). This is used to stop displaying a message at a certain day/time.
 - **featureImageUrl**: *(optional)* Used by popup announcements and announcements on the Features page.
 - **priority**: Set to "high" if you want the message to be displayed with higher visibility (i.e. As a priority notification or popup announcement, respectively).
+- **recurrence**:(optional) If true, even if a notification is dismissed, it will continue to reoccur in the user's home at the start of every session until the user is no longer a member of the targeted group. For example, if a user is a member of students-with-outstanding-parking-tickets, that user will be confronted with the notification at every login until they pay the fine.
```
apetro: If this is a separate field, let's call it out as, like `dismissible`, "experimental", so that we remember why we don't have to semantically version evolutions of it for now.

PR #592, opened 2017-10-31
fix: refactor time-sensitive widget

davidmsibley:
Reworking the time-sensitive-content widget logic. The bug of not reverting back to a basic widget post-templateRetireDate was largely fixed with the change in the partial. It was happening because we were getting into a place where `showTimeSensitiveContent == false` and `showBasicContent == false`

That's been replaced by a check for the presence of a callToAction. The rest of the refactor revolves around using actual `Date`s in the logic, rather than the strings, and breaking decision making blocks out into unit-test-_able_ functions.

`resolveTemplateStatus` smells a little funky. It's basically a bunch of nested ifs, or a backwards switch statement. I think it reads a little clearer like this, but I'd be ok refactoring it as a switch.

----

Contributor License Agreement adherence:

- [x] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

[Apereo CLA roster]: http://licensing.apereo.org/completed-clas
[Contributor License Agreements]: https://www.apereo.org/licensing/agreements

PR #593, opened 2017-11-01
Moment.js Refactor

ChristianMurphy:
Reworking the time-sensitive-content widget logic. The bug of not reverting back to a basic widget post-templateRetireDate was largely fixed with the change in the partial. It was happening because we were getting into a place where `showTimeSensitiveContent == false` and `showBasicContent == false`

That's been replaced by a check for the presence of a callToAction. The rest of the refactor revolves around using actual `Date`s in the logic, rather than the strings, and breaking decision making blocks out into unit-test-_able_ functions.

`resolveTemplateStatus` smells a little funky. It's basically a bunch of nested ifs, or a backwards switch statement. I think it reads a little clearer like this, but I'd be ok refactoring it as a switch.

---

refactor of #592 with moment js.

----


ChristianMurphy:
:notebook: Codeclimate is flagging `TODO` comments left for future ES2016+ refactoring

components/portal/widgets/controllers.js
```diff
@@ -457,53 +459,55 @@ define(['angular'], function(angular) {
...
 
-      if (now && actionStart && actionEnd) {
-        angular.forEach(statuses, function(status) {
-          if (status.check()) {
-            result = status.name;
-          }
-        });
-      }
+      var result = '';
+      // TODO: rewrite as Array.prototype.find when IE support is dropped
```
davidmsibley: note, this isn't a drop in replacement.  If you write it with `find` the order of the statuses array should be reversed, as 'lastDay' is a subset of 'ongoing', so it needs to be found first. It's currently written to fall through to more stringent classifications.

components/portal/widgets/controllers.js
```diff
@@ -457,53 +459,55 @@ define(['angular'], function(angular) {
     };
 
     var resolveTemplateStatus = function(now, actionStart, actionEnd) {
-      var result = '';
+      if (!now || !actionStart || !actionEnd) {
+        throw new TypeError('now, actionStart, and actionEnd must all be set');
```
davidmsibley: I'd rather not throw errors. Especially if they don't get caught.

If a widget is misconfigured it now would break the page, instead of just failing back to the basic widget view like it does in every other failure scenario.

components/portal/widgets/controllers.js
```diff
@@ -457,53 +459,55 @@ define(['angular'], function(angular) {
...
-      var result = -1;
-      var oneDay = 1000 * 60 * 60 * 24;
-
-      if (now && actionEnd && now < actionEnd) {
-        var difference = actionEnd.getTime() - now.getTime();
-        result = Math.ceil((1.0 * difference) / oneDay);
-      }
-
-      return result;
+      return moment(actionEnd).diff(now, 'days');
```
davidmsibley: I'd want to check to make sure this behavior matches the existing behavior: when there's less than a day left, it should report 1 day.

components/portal/widgets/controllers.js
```diff
@@ -414,41 +414,43 @@ define(['angular'], function(angular) {
...
+        return fullISODate.toDate();
+      }
+
+      // if it is a partial ISO string, fill in missing fields
+      var partialISODate = moment(str, ['YYYY-MM-DD', 'MM-DD'], true);
+      if (partialISODate.isValid()) {
+        if (defaultEndOfDay) {
+          return partialISODate.endOf('day').toDate();
+        } else {
+          return partialISODate.startOf('day').toDate();
```
davidmsibley: please don't have multiple returns in a function. define the base case, then specify the outliers.

components/portal/widgets/controllers.js
ChristianMurphy: Very true, luckily there is an `Array.prototype.reverse` function that can handle reorder nicely.
Or it could just be reversed in the code :wink:

components/portal/widgets/controllers.js
ChristianMurphy: I feel like we've had this discussion before.
Single return makes a lot of sense when dealing with side effects.
multiple returns or early return reduces complexity when dealing with pure functions.
This is a pure function, multiple returns makes sense.

components/portal/widgets/controllers.js
ChristianMurphy: This doesn't exactly match that behavior, fix incoming

components/portal/widgets/controllers.js
ChristianMurphy: Semantically throwing a `TypeError` is correct here, however if it will break the app, it can be dropped to a blank `return`

apetro: You're both right. The immediate function has constraints on its arguments. Passing it arguments that don't meet those constraints, that's an error. But if this function is going to error, anything that calls this needs to appropriately handle that error, so that errors in widgets don't escape that specific widget to wreck the whole page.

components/portal/widgets/controllers.js
davidmsibley: yes, we have, but I'm having a difficult time seeing the practical advantage. Was the function not "pure" with a single return statement? There were no side effects, the only mutations were scoped within the function. The two implementations are almost the same in terms of complexity: http://jsmeter.info/cs6kwf/1

With multiple return statements, the code is less malleable and bugs take longer to diagnose. The bug that precipitated this refactor was itself a result of having multiple `showBasicWidget()` paths, one `displayTimeSensitiveContent()` path, and a not immediately noticeable path where neither of them were run. To see that, I had to trace through the code to find that case, and not rely on the assumption that the code was hitting one of those prior cases.  Your function ends with a return of an invalid date, which would be an acceptable default, if any of the other code actually checked `isValid`. That aside, how quickly would you notice if that last return happened to be missing? With three other returns in the function, it wouldn't be immediately noticeable. If there's only one return statement, you know immediately if it's incorrect.

I'm not bringing this up to be contrarian. I'm trying to optimize for resiliency and ease of maintenance, and as of yet I don't see how multiple return statements do that.

ChristianMurphy: > Was the function not "pure" with a single return statement?

it is pure either way.

> With multiple return statements, the code is less malleable and bugs take longer to diagnose. 

I agree that quick diagnostic and malleability are good.
Multiple returns are more malleable. Single return tends to cause deeply nested conditions that are a nightmare to debug. Multiple returns encourages flat code, with conditions at the top level.

> bugs take longer to diagnose

How? Either way the first thing a developer should do is add a breakpoint and step through the code.
With multiple returns there is less complexity to debug.

> if any of the other code actually checked `isValid`

Why check `isValid`?
`moment` handles invalid automatically.
https://momentjs.com/docs/#invalid-moments

> at aside, how quickly would you notice if that last return happened to be missing?

Pretty quickly, Chrome dev tools highlight what ever value in a scope is.
Step through would catch this pretty quick.

> If there's only one return statement, you know immediately if it's incorrect.

Value is `underfined` there wasn't a `return`, multiple returns debugging seems pretty instant to me.

> I'm trying to optimize for resiliency and ease of maintenance

Same. Again agree philosophically, but differ in implementation.

> and as of yet I don't see how multiple return statements do that.

I've used single `return` in the past.
Deep conditionals that are hellish to debug tend to build up overtime.
Keeping flat conditionals using multiple `return`s and `throw`s avoids that.

davidmsibley:
Shoot, sorry. I used "Rebase and merge" on my pull request instead of just a merge, which I think wouldn't have caused the conflicts.

ChristianMurphy:
> Shoot, sorry. I used "Rebase and merge" on my pull request instead of just a merge, which I think wouldn't have caused the conflicts.

No worries, a simple rebase off `master` resolves the conflicts.

> I read up on moment and love it.
> +1 for using moment.

Glad you like it!
At Node Interactive the Moment team announced that there is a proposal in to TC39 to use Moment as a reference implementation for new unitTime and relativeTime APIs for ES2019+.

> I'm not a big fan of this `resolveDate` function, though it's hard for me to put exactly what it is. I think we need to refactor the logic. [...] This becomes a pretty simple online moment call then. Now typing, I am just grumpy about this logic having to be needed.

I'm open to suggestions. I simply ported over the existing logic.
For making more dramatic changes I would probably need to have a live API with real data to test against.

> I propose stronger documentation stating that behavior for dates is that you provide an ISO date with time stamp, or it defaults to the beginning of the day.

:+1: that would likely fallout side the scope of this current refactor.

> Also, I generally don't like TODO's. I feel like they just hang around until a refactor happens anyways. I'm all for GitHub issues though.

:man_shrugging: Okay, the GitHub issue would just hang around until a refactor happens as well.

components/portal/widgets/controllers.js
ChristianMurphy: Updated in a311bb0

components/portal/widgets/controllers.js
```diff
@@ -414,41 +414,43 @@ define(['angular'], function(angular) {
...
       // Check if today falls within any of the provided date ranges
       angular.forEach(callsToAction, function(callToAction, index) {
         if (result === -1) { // We haven't found a call to action
           var startDate =
             resolveDate(callToAction.activeDateRange.templateLiveDate);
           var endDate =
             resolveDate(callToAction.activeDateRange.templateRetireDate, true);
-          if (startDate && startDate < now // Template started
-            && (!endDate || endDate > now)) { // Template has not ended
+          if (moment().isBetween(startDate, endDate)) {
```
davidmsibley: This changed the behavior a little bit. If an endDate isn't specified, this comes up false. That's fine, that's how the previous code worked, but if we keep it like this, should change the documentation back

components/portal/widgets/controllers.js
```diff
@@ -457,53 +459,56 @@ define(['angular'], function(angular) {
     };
 
     var resolveTemplateStatus = function(now, actionStart, actionEnd) {
-      var result = '';
+      if (!now || !actionStart || !actionEnd) {
```
davidmsibley: This is never true now, actionStart and actionEnd are invalid instead of null if they're missing from the config

davidmsibley:
Try running this locally. To do that, I `mvn clean install` uportal-app-framework, then pull down and change widget-creator to pull in the SNAPSHOT version, `mvn clean package` then `mvn jetty:run`

components/portal/widgets/controllers.js
ChristianMurphy: updated

ChristianMurphy:
Changes have been rebased off `uPortal-Project:master` branch and fix for invalid date detection has been added.

components/portal/widgets/controllers.js
thevoiceofzeke: Is this related to an outdated commit?

davidmsibley: nope. this line works like the pre- #592 code, in that it requires a `templateRetireDate` if there's a `templateLiveDate`. #592 changed that such that `templateRetireDate` was optional even if a `templateLiveDate`, and edited the documentation to say that.  Line 452 on the left hand side is checking whether the template has been retired, rather than whether now is between two valid dates.

This comment was just a poorly worded reminder to edit the documentation when changes in functionality occur.

PR #595, opened 2017-11-02
fix: Broadcast changes to variables needed in many places

thevoiceofzeke:
**In this PR**:
- Stop main-menu from scrolljacking user when opened
- Use messages service to broadcast when the existence of priority notifications and unseen announcements has changed
- Add listeners to the controllers that need to know about those events
- Stop passing controllers as attributes of directives

This is all needed mainly to allow the side navigation to work as expected. Without this, the sidenav automatically forces scroll-to-top when opened. It's a buggy behavior that has been ignored by Angular Material devs, despite a number of issues being opened. This is a workaround. 
 
The added benefit is that several directives need to know when/if there are priority notifications or unseen announcements (e.g. the `notifications-bell` and `main-menu` directives). Because of the DOM hierarchy, passing controllers as directive scope variables no longer works in all cases (and is a pretty hack thing to do in the first place). Using a service to broadcast changes to those variables is arguably cleaner and easier to understand.

### Demo

scroll-demo

----

### DIFF
--- a/components/css/buckyless/directives/main-menu.less
+++ b/components/css/buckyless/directives/main-menu.less
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-main-menu {
+main-menu:not(.push-content) {
   .main-menu__mobile-bar {
     background: @grayscale2;
     min-height: 56px;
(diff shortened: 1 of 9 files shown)


thevoiceofzeke:
Closing due to missed bugs.

PR #597, opened 2017-11-03
docs: Recommend title case not sentence case for list-of-links labels

apetro:
In prototyping recent list-of-links usage with UX consultant @mrdahman , title case rather than sentence case seemed to be preferable for the within-`list-of-links`-type-widget link labels.

star-widget-prototype

This is at odds with the style recommendation currently embedded in the widget documentation (though interestingly the example screenshot in the documentation is already using title case rather than sentence case for this labelling). This changeset would update the guidance to match what we seem to now be preferring.

Personally, I'd have gone with sentence case. But this isn't about me personally. :) So this PR is about reconciling what professionally we ought to be recommending.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased][]
 
+* Recommend title case rather than sentence case for links in list-of-links widgets
+
 ## [6.1.0][] - 2017-10-18
 
 ### Added
--- a/docs/make-a-widget.md
+++ b/docs/make-a-widget.md
@@ -122,13 +122,13 @@ Follow these steps for each of the predefined widget types described in this doc
       "launchText":"Launch talent development",
       "links": [
         {
-          "title":"All courses and events",
+          "title":"All Courses and Events",
           "href":"https://www.ohrd.wisc.edu/home/",
           "icon":"fa-at",
           "target":"_blank"
         },
         {
-          "title":"My transcript",
+          "title":"My Transcript",
           "href":"https://www.ohrd.wisc.edu/ohrdcatalogportal/LearningTranscript/tabid/57/Default.aspx?ctl=login",
           "icon":"fa-envelope-o",
           "target":"_blank"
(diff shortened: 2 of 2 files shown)


mrdahman:
I think a call to action looks better as title case. Sentence case without punctuation at the end is a little annoying to me personally, but more than that, I think it just looks more impressive. I admit I'm also biased by German immersion, which capitalizes liberally. :)

prtreige:
I don't want to throw a wrench in @mrdahman reco's but I'm still an advocate for sentence case, per Material Guidelines, and though it's hardly settled in terms of usability research, sentence case has strong advantages for read-ability. See https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98.

apetro:
I think this is the right soul-searching to be prompting, albeit over an elongated time period so that @mrdahman can enjoy her justly earned leave this week and re-engage next week or so. :)

Stipulating that our current realization of Material Design is imperfect and as much aspirational as it is achieved, this is another one of those opportunities to decide: are we literally adopting Material Design, as best we can, and of course layering on local considerations and working within its extensibility for typography and branding? Or are we working towards our own design system somewhat inspired by Material Design? I see efficiencies, leverage to be gained in the former.

mrdahman:
That's a great point @prtreige . Readability is a concern with title case, though somehow in the specific example of the widget we were looking at, the two documents that we put in title case were easier to understand in title case than sentence case. Perhaps in that instance it was more of an example of a proper noun, i.e., a named document. (Also, as an aside, the character limit on the list of links widgets doesn't allow us to write anything that would be too difficult to read in title case.) But yes, I agree that we run into the larger problem of how closely we are adhering to Material Design. And, @apetro , I fully acknowledge that there are efficiencies with literally adopting Material Design, but think there might also be some wiggle room if something feels easier to understand. Thoughts?

apetro:
Hard cases make bad law.

I think the most important error here was my prematurely trying to generalize a change to the design system from one goofy `list-of-links` widget making exceptions to the design system. Too soon.

Let's 

+ revert this specific changeset and drop the `search-with-links` parallel in #597 ,
+  keep the guidance we had (sentence case) (think of it as a nod to *stare decisis*)
+ accept the MyUW-local one-off ad-hoc title case usage in the one specific goofy widget because it's literally what we've been asked to do locally, temporarily accepting a little local MyUW weirdness to shield the open source product from turbulence
+ work more generally to really grapple with what our relationship to Material Design is and ought to be
+ revisit this when it comes up again. I imagine it will.

PR #603, opened 2017-11-10
fix: Add index tracking to notifications page

thevoiceofzeke:
#602 fixed the bug, but we need the same fix on the notifications page.

*Note: This bug seems to have been caused by two notifications that have identical IDs -- in reality we should never encounter this issue as we are the gatekeepers of production notifications.*

----


Doug-Reed:
👍

PR #604, opened 2017-11-11
Catch up changelog for putative 7.0.0 release

apetro:
Catches up `CHANGELOG.md` to document the changes pending release, including an important breaking change, positing that the next release will therefore be a "major" release versioned `7.0.0`.

This makes the changelog more current, but probably is not sufficient documentation of just what the nature of the breaking change is and what upgraders should do about it when they upgrade.

----


apetro:
Apparently the new latest failure is because the `scope` portion of the `style(CHANGELOG): indent child bullets with 4 not 2 spaces` commit message is not lowercase but must be. I'll see about rewriting history enough to fix that.

Issue #607, opened 2017-11-14
Fail gracefully when messagesURL not configured


apetro:
Setting `SERVICE_LOC.messagesURL` to `null` or `""` should be valid configuration meaning "turn off the messages features in this app". No notifications, no mascot announcements.

In practice this configuration currently breaks `uportal-app-framework` apps.

apetro:
Observation: maybe not configuring `messagesURL` should be the default state in the app framework. messages is an integration that one connects up in the course of deploying an app into, integrating it into a holistic "portal". When locally developing, it might be best if the messages features were out of the way so as to not distract.

apetro:
This tracks publicly an issue MyUW tracks privately as [MUMUP-3188|https://jira.doit.wisc.edu/jira/browse/MUMUP-3188] for local Scrum purposes.

PR #611, opened 2017-11-14
fix: Fail gracefully when messagesURL not configured

thevoiceofzeke:
[MUMUP-3188](https://jira.doit.wisc.edu/jira/browse/MUMUP-3188): "See #607. Setting SERVICE_LOC.messagesURL to null or an empty string should result in the bell not appearing. The current behavior is that null results in a broken build and an empty string result in an error message. This has to be changed to frame apps can choose to not show any notifications."

**In this PR:**
- Don't crash build if `messagesURL` is set to `null`
- Hide notifications/announcements features if `messagesURL` is set to `null` or empty string

----


thevoiceofzeke:
Is there a softer "fail" for codeclimate? It really hurts me to see that red x on my pull requests because of "similar code".

components/js/override.js
```diff
@@ -22,13 +22,10 @@ define(['angular'], function(angular) {
...
-          {
-            'label': 'MyUW home',
-            'icon': 'home',
-            'url': '/web',
-          },
-        ],
+        'enablePushContentMenu': true,
+      },
+      'SERVICE_LOC': {
+        'messagesURL': 'staticFeeds/sample-messages.json',
```
apetro: Maybe the ootb `override.js` should not override the default messages-are-off `messagesURL` to keep this feature out of the way of local app development (which most of the time won't be developing messages content).

components/portal/main/controllers.js
```diff
@@ -182,6 +183,7 @@ define(['angular', 'require'], function(angular, require) {
       vm.hideMainMenu = false;
       vm.footerLinks = FOOTER_URLS;
       vm.openMenuByDefault = false;
+      vm.hideMessagesFeatures = false;
```
apetro: While parallel with `vm.hideMainMenu`, naming the variable "hide" has a double-negative effect. "hide=false" means the feature is turned on. It might be better to use a variable that states the feature positively, as in "vm.messagesFeature" being true when the feature is turned on and false when it is off/hidden.

Notice this play out in the templates, with `&& !vm.hideMessagesFeatures` for "and if the feature is on".

components/portal/messages/controllers.js
```diff
@@ -128,8 +128,16 @@ define(['angular'], function(angular) {
...
           $scope.messages = {};
 
-          if (!$rootScope.GuestMode) {
+          if (!$rootScope.GuestMode && SERVICE_LOC.messagesURL
+            && SERVICE_LOC.messagesURL !== '') {
             getMessages();
+          } else {
+            // Messages features aren't configured properly, possibly
+            // on purpose. Communicate this and hide features.
+            $scope.hideMessagesContent = true;
```
apetro: Prefer positive variables? less hide-is-true and more show-is-false?

apetro:
Didn't see settings to adjust for CodeClimate pickiness. Removed `uportal-app-framework` from CodeClimate participation to reduce the hurt.

PR #613, opened 2017-11-15
fix: Sensible aria labels for widgets, display bugs

### LINKED ISSUE #608: aria-labels for widget launch buttons
As a screen reader user, I would like launch buttons to have labels that are meaningful considered in isolation, so that I can exercise widgets effectively by stepping through the hyperlinks on the page.

Currently problematic examples:

+ For a `list-of-links` widget configured with launch bar text of `"Launch full app" per guidance, `aria-label` is also `"Launch full app"` rather than something that would help the user understand what app will be launched.
+ For a `search-with-links" widget configured with launch bar text of `"Open website"` per guidance, `aria-label` is also `"Open website"` rather than something that would help a user understand what website will be opened.

thevoiceofzeke:
[MUMUP-3146](https://jira.doit.wisc.edu/jira/browse/MUMUP-3146): "As a screen reader user, I would like widget launch buttons to have labels meaningful considered in isolation, so that I can exercise widgets effectively by stepping through the hyperlinks on the page."

**In this PR**:
- Resolves #608 
- Predefined widget types have launch button aria text of "Launch [widget title]"
- Updated docs to make guidance more visible for custom widgets
- Set redundant widget title to aria-hidden (title is included in the widget's first element, prior to description) to improve screen reader flow
- Continue flex-based model for push-content side navigation
- Move horizontal overflow limit to highest element (document body) to cope with potentially poorly-designed app content areas while using push-content navigation

----

### DIFF
--- a/components/js/override.js
+++ b/components/js/override.js
@@ -27,5 +27,8 @@ define(['angular'], function(angular) {
       'SERVICE_LOC': {
         // 'messagesURL': 'staticFeeds/sample-messages.json',
       },
+      'MESSAGES': {
+        // 'notificationsPageURL': '/notifications',
+      },
     });
 });
--- a/components/js/override.js
+++ b/components/js/override.js
@@ -25,7 +25,7 @@ define(['angular'], function(angular) {
         'enablePushContentMenu': true,
       },
       'SERVICE_LOC': {
-        'messagesURL': 'staticFeeds/sample-messages.json',
+        // 'messagesURL': 'staticFeeds/sample-messages.json',
       },
       'MESSAGES': {
         // 'notificationsPageURL': '/notifications',
--- a/components/js/override.js
+++ b/components/js/override.js
@@ -25,7 +25,7 @@ define(['angular'], function(angular) {
         'enablePushContentMenu': true,
       },
       'SERVICE_LOC': {
-        // 'messagesURL': 'staticFeeds/sample-messages.json',
+        'messagesURL': 'staticFeeds/sample-messages.json',
       },
       'MESSAGES': {
         // 'notificationsPageURL': '/notifications',
--- a/components/portal/messages/controllers.js
+++ b/components/portal/messages/controllers.js
@@ -86,16 +86,15 @@ define(['angular'], function(angular) {
                 messagesService.getMessagesByTitle(allMessages),
             };
           }
-            // Call filtered notifications promises, then pass on to
-            // the completion function
-            $q.all(promiseFilteredMessages)
-              .then(filterMessagesSuccess)
-              .catch(filterMessagesFailure);
-         
-            // Separate all messages by their types
-            $scope.messages = $filter('separateMessageTypes')(allMessages);
-            // Change hasMessages so child controllers can pick up on it
-            $scope.hasMessages = true;
+          // Call filtered notifications promises, then pass on to
+          // the completion function
+          $q.all(promiseFilteredMessages)
+            .then(filterMessagesSuccess)
+            .catch(filterMessagesFailure);
+          // Separate all messages by their types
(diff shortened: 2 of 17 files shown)


components/portal/messages/controllers.js
```diff
@@ -86,16 +86,15 @@ define(['angular'], function(angular) {
...
-              .then(filterMessagesSuccess)
-              .catch(filterMessagesFailure);
-         
-            // Separate all messages by their types
-            $scope.messages = $filter('separateMessageTypes')(allMessages);
-            // Change hasMessages so child controllers can pick up on it
-            $scope.hasMessages = true;
+          // Call filtered notifications promises, then pass on to
+          // the completion function
+          $q.all(promiseFilteredMessages)
```
ChristianMurphy: This works :white_check_mark: 
But this is an area where `$q` diverged from `Promise`. In the `Promise` spec, `all` only supports iterable. [ref 1] Plain Object is not considered an interable. [ref 2] Should there be a plan to upgrade to ES2015+ this would be an area to flag as a risky upgrade point.

[ref 1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all
[ref 2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol

components/portal/messages/services.js
```diff
@@ -208,6 +208,61 @@ define(['angular'], function(angular) {
...
+         * @param {Array} messages An array of message objects
+         * @return {filteredMessages[]} an array of messages that
+         *   includes messages that requested a title from a url
+         */
+        var getMessagesByTitle = function(messages) {
+          var promises = [];
+          var filteredMessages = [];
+          var toBeDiscarded = [];
+          angular.forEach(messages, function(message) {
+            if (message.titleUrl) {
```
ChristianMurphy: :art: If possible prefer `Array.prototype.filter` for messages with a `titleUrl` to reduce complexity of `forEach`

components/portal/messages/services.js
```diff
@@ -208,6 +208,61 @@ define(['angular'], function(angular) {
...
+              promises.push($http.get(message.titleUrl)
+                .then(function(result) {
+                  if (result.data) {
+                    var titleObject = result.data;
+                    if (titleObject.result && titleObject.result.length > 0) {
+                      var fromApi = titleObject.result[0];
+                      if (fromApi.full) {
```
ChristianMurphy: I'm not fond of the idea of four nested `if`s, but I see why there are here. :white_check_mark: 
If, in the future, `lodash` were added `lodash.get` could abstract some of this complexity away.

components/js/override.js
```diff
@@ -25,7 +25,10 @@ define(['angular'], function(angular) {
         'enablePushContentMenu': true,
       },
       'SERVICE_LOC': {
-        // 'messagesURL': 'staticFeeds/sample-messages.json',
+        'messagesURL': 'staticFeeds/sample-messages.json',
```
apetro: `override.js` `SERVICE_LOC.messagesURL` changes in this changeset from [the change](https://github.com/uPortal-Project/uportal-app-framework/pull/611/files#diff-46d54ea0f3ac49e925d8968b0e250c2dR28) in #661 why?

thevoiceofzeke: Switched it on during the tedium of resolving linting errors, forgot to switch it off.

components/portal/messages/services.js
thevoiceofzeke: We're hoping to move most of this filtering logic to a microservice in the near-ish future, so hopefully the front end won't have to bother with it at all.

PR #620, opened 2017-11-21
MUMMNG-3983 - Synch header bar with priority notification bell

Doug-Reed:
Dismissing a priority notification in the bell did not cause the header bar to update. There are two instances of the controller running at the same time. Call to $rootScope to cause a refresh of the priority notification count. 

----

### DIFF
--- a/components/portal/messages/controllers.js
+++ b/components/portal/messages/controllers.js
@@ -193,11 +193,17 @@ define(['angular'], function(angular) {
             // Else hide messages features
             if (angular.equals($scope.$parent.showMessagesFeatures, true)) {
               configureNotificationsScope();
+              checkForChangesInOtherInstances();
             } else {
               vm.showMessagesFeatures = false;
               vm.isLoading = false;
             }
           }
+        });
+
+        var checkForChangesInOtherInstances = $rootScope.$watch(
+          'priorityNotificationCount', function() {
+            configurePriorityNotificationsScope();
         });
 
         // ////////////////
@@ -284,11 +290,18 @@ define(['angular'], function(angular) {
          * Alert the UI to show priority notifications if they exist
          */
         var configurePriorityNotificationsScope = function() {
+          var howManyPriorities = $rootScope.priorityNotificationCount;
           // Use angular's built-in filter to grab priority notifications
           vm.priorityNotifications = $filter('filter')(
             vm.notifications,
             {priority: 'high'}
           );
+
+          if (vm.priorityNotifications.length != howManyPriorities) {
+              $rootScope.priorityNotificationCount
+                  = vm.priorityNotifications.length;
+            }
+
           // If priority notifications exist, notify listeners
           messagesService.broadcastPriorityFlag(
             vm.priorityNotifications.length > 0
@@ -351,6 +364,7 @@ define(['angular'], function(angular) {
           if (isHighPriority) {
             clearPriorityNotificationsFlags();
           }
+          configurePriorityNotificationsScope();
         };
 
         /**


components/portal/messages/controllers.js
```diff
@@ -205,13 +205,19 @@ define(['angular'], function(angular) {
             // Else hide messages features
             if (angular.equals($scope.$parent.showMessagesFeatures, true)) {
               configureNotificationsScope();
+              checkForChangesInOtherInstances;
```
ChristianMurphy: :question: does this need parenthesis to be called?

components/portal/messages/controllers.js
```diff
@@ -205,13 +205,19 @@ define(['angular'], function(angular) {
...
             } else {
               vm.showMessagesFeatures = false;
               vm.isLoading = false;
             }
           }
         });
 
+        var checkForChangesInOtherInstances = $rootScope.$watch(
+          'priorityNotificationCount', function() {
+          configurePriorityNotificationsScope();
```
ChristianMurphy: :art: indent content within a function scope one level in from the function.

components/portal/messages/controllers.js
Doug-Reed: The committer yelled at me when I didn't assign the $rootScope.$watch to a variable, then it yelled at me when I assigned it to a variable, but didn't reference it.

Referencing it in line 208 made my problems go away. 

var checkForChangesInOtherInstances = $rootScope.$watch(
          'priorityNotificationCount', function() {
            configurePriorityNotificationsScope();
        });

components/portal/messages/controllers.js
Doug-Reed: Done

components/portal/messages/controllers.js
ChristianMurphy: Thats coming from https://github.com/Gillespie59/eslint-plugin-angular/blob/master/docs/rules/on-watch.md

the variable is to remove the watcher registration.
if the watcher is never unregistered that's a potential red flag.

components/portal/messages/controllers.js
```diff
@@ -205,13 +205,19 @@ define(['angular'], function(angular) {
...
               configureNotificationsScope();
+              checkForChangesInOtherInstances;
             } else {
               vm.showMessagesFeatures = false;
               vm.isLoading = false;
             }
           }
         });
 
+        var checkForChangesInOtherInstances = $rootScope.$watch(
```
thevoiceofzeke: I'd like to suggestusing a service and `$broadcast` that we have in place already instead of adding more variability to the notifications logic that I already made very complicated.

Take a look at the [messagesService](https://github.com/uPortal-Project/uportal-app-framework/blob/master/components/portal/messages/services.js#L337). We can leverage that to add a number (count) to what we're broadcasting so that the same listener we're using in the [main menu](https://github.com/uPortal-Project/uportal-app-framework/blob/master/components/portal/main/controllers.js#L206) can be reused in the priority notifications.

Since you just left for vacation, I would be willing to submit a separate PR with that model if you want.

Doug-Reed: The problem is that there are two instances of that controller, and they don't know about each other - even when the common service is invoked.

apetro: Do the controllers need to know about each other? Some kind of pub-sub pattern might work so that each controller only needs to know about the (singleton) service, the controllers call the service to make changes, and the service publishes the updated state to subscribed controllers?

This is one of those moments when the best way to talk about code is with code. That separate PR from @thevoiceofzeke would be interesting and we'd learn from it whether we ended up adopting it or not.

Doug-Reed:
Superseded by a later and cleaner pull request - #638

PR #629, opened 2017-11-29
style: wrap recent changelog at 80 chars per line

apetro:
Wrap recent changelog lines at 80 characters per line.

----


apetro:
If only there were some way this linting could automatically run for me when I locally commit... :)

apetro:
(That next failure was because my commit message on the linting fix ended with a period and so itself failed commit message linting. If only there were some way my commit messages could be linted locally when I commit them... :) ).

So, the helpful version of this joke is this: that there are tradeoffs between local and CI linting, and I'm navigating them, badly.  One root problem here is that this linting isn't working for me locally -- I imagine `npm run lint-md` is supposed to let me lint my Markdown locally, but locally my latest error message when I try to run that is `Cannot find module 'trough'`.

vertein:
Ahh, the old git commit messages aren't sentences gotcha.  

Have you tried npm install before running npm run lint-md?  I'll pull down the code and see if it works for me.

apetro:
So, yes, I had tried `npm install`, twice. Tried it one more time now. And of course now `npm run lint-md` works just fine for me locally.

apetro:
(And apparently the next failure was because my commit message led with `4` after the scope. This will eventually be funny. Squashed so I only have to get 1 commit message right.)

ChristianMurphy:
Hey @apetro :wave: 
Chiming in a bit late here.
You may enjoy using commitizen `npm run cz` or commitlint prompt `npm run commit`.
Both provides a commandline WYSIWYG for crafting conventional commit messages.


Also, it may be valuable if you share your experiences and thoughts on how to improve the commit message experience at https://github.com/marionebl/commitlint/issues/94.

ChristianMurphy:
> So, yes, I had tried npm install, twice. Tried it one more time now. And of course now npm run lint-md works just fine for me locally.
 
I'd check which version of npm is being run versions 5.1.0+ are much more consistent with installing modules than previous versions.
ref: https://docs.npmjs.com/troubleshooting/try-the-latest-stable-version-of-npm

Its also worth noting that running `npm install` multiple times can lead to some non-determinism in the structure/versions in the *node_modules* folder.
For the most consistent results `rm -rf node_modules && npm install`.
ref: https://docs.npmjs.com/how-npm-works/npm3-nondet

ChristianMurphy:
> If only there were some way my commit messages could be linted locally when I commit them

That's the dream. :slightly_smiling_face: 
Its worth noting that the webhooks removed as part of the uPortal-Project/uportal-app-framework/issues/535 purge of githooks, can still be manually added and maintained within a local setup of uPortal-app-framework.
It just take more effort and initiative.

PR #636, opened 2017-12-04
build(lint-staged): suppress eslint on `docs/`

apetro:
Since `lint-js` ignores docs, also ignore docs when JS-linting staged changes.

(I ran into this when touching `app-options.md`: while I didn't touch the example JavaScript, the example JavaScript failed `eslint`.)

Arguably we should be linting the JavaScript embedded in Markdown files -- but if we should, we should do so consistently.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 + Notes adopting Conventional Commits message style in `CONTRIBUTING.md`, with
 tips for easing compliance. (#634)
 
+### Build
+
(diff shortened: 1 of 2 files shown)


vertein:
> I agree that the ideal solution is to check JS files/examples in docs/
I agree

> Arguably we should be linting the JavaScript embedded in Markdown files -- but if we should, we should do so consistently.
I agree

apetro:
I'd support a changeset that turns back on linting JS examples embedded in Markdown docs, so long as that changeset comes with the fixes to make those JS examples succeed at linting, and turns on this linting for all the ways of triggering Markdown linting so it's less easy to get confused about whether a given `.md` file is passing linting. 😄  Captured as #637 .

PR #639, opened 2017-12-12
MUMUP-3148: Messages filtered by start and end dates

Doug-Reed:
Adding a filter to weed out any messages which fall outside the bounds of the goLiveDate or expireDate
----


components/portal/messages/filters.js
```diff
@@ -67,6 +67,40 @@ define(['angular'], function(angular) {
           }
         });
       };
+    })
+     .filter('filterMessagesWithInvalidDates', function() {
+      var retVal = [];
+      return function(messages) {
+          var thePresent = Date.now();
+          angular.forEach(messages, function(message) {
```
ChristianMurphy: :art: prefer `map` over `forEach + retVal + push` 

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map
ref: https://code.tutsplus.com/tutorials/how-to-use-map-filter-reduce-in-javascript--cms-26209
ref: https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/prefer-map.md

components/portal/messages/filters.js
```diff
@@ -67,6 +67,40 @@ define(['angular'], function(angular) {
...
+            if (message.expireDate || message.goLiveDate) {
+              var addToRetVal = true;
+              if (message.goLiveDate) {
+                var goDate = new Date(message.goLiveDate);
+                var goTime = goDate.getTime();
+                if (thePresent < goTime) {
```
ChristianMurphy: :art: prefer [`moment.isBefore`](https://momentjs.com/docs/#/query/is-before/) over `Date` -> `Number` conversion, followed by compare.

components/portal/messages/filters.js
thevoiceofzeke: It occurs to me we shouldn't be using `map` because we're not trying to transform/return every element. Rather, we're trying to filter out elements that don't match certain criteria, so wouldn't `Array.filter` be more appropriate? See #641

ChristianMurphy: Ah, I misread the data flow, `filter` would be more appropriate in this case.

components/portal/messages/filters.js
```diff
@@ -67,6 +67,40 @@ define(['angular'], function(angular) {
           }
         });
       };
+    })
+     .filter('filterMessagesWithInvalidDates', function() {
+      var retVal = [];
```
davidmsibley: Let's move the `retVal` a line down into the returned function, rather than using it as a closure.  Right now if you were to do `var foo = $filter('filterMessagesWithInvalidDates')` and then execute foo twice, the second time you'd get the array with double the entries.

Stylistically we never do that, but why risk it.

Doug-Reed:
Filter is refactored as an array.filter using moment. Take another look and see if this'll work.

components/portal/messages/filters.js
```diff
@@ -68,6 +68,31 @@ define(['angular'], function(angular) {
...
+              if (goDate.isBefore(thePresent)
+                && stopDate.isAfter(thePresent)) {
+                return message;
+              }
+            }
+            if (goDate.isValid() && goDate.isBefore(thePresent)) {
+              return message;
+            }
+            if (stopDate.isValid() && stopDate.isAfter(thePresent)) {
+              return message;
```
davidmsibley: I'm not a fan of multiple return statements, but prior experience indicates I'm outnumbered.

components/portal/messages/filters.js
```diff
@@ -68,6 +68,31 @@ define(['angular'], function(angular) {
...
+          var thePresent = moment.now();
+          return messages.filter(function(message) {
+            var goDate = moment(message.goLiveDate);
+            var stopDate = moment(message.expireDate);
+            // If neither date is populated, message is valid.
+            if (!goDate.isValid() && !stopDate.isValid()) {
+              return message;
+            }
+            if (goDate.isValid() && stopDate.isValid()) {
+              if (goDate.isBefore(thePresent)
```
thevoiceofzeke: Moment has the [isBetween()](https://momentjs.com/docs/#/query/is-between/) function for this kind of check. Makes no difference what you use, but it's out there!

components/portal/messages/controllers.js
```diff
@@ -103,8 +103,10 @@ define(['angular'], function(angular) {
               result.filteredByGroup,
               result.filteredByData
             );
+            var dateFilteredMessages =
+              $filter('filterMessagesWithInvalidDates')(filteredMessages);
```
thevoiceofzeke: Semantic gripe: This filter isn't actually filtering things with invalid dates (i.e. If there aren't dates specified, those dates are invalid, but those messages aren't filtered out).  Would prefer a name like `filterByDates`.

¯\\\_(ツ)_/¯

components/portal/messages/filters.js
```diff
@@ -68,6 +68,31 @@ define(['angular'], function(angular) {
...
       };
     })
+    .filter('filterMessagesWithInvalidDates', function() {
+      return function(messages) {
+          var thePresent = moment.now();
+          return messages.filter(function(message) {
+            var goDate = moment(message.goLiveDate);
+            var stopDate = moment(message.expireDate);
+            // If neither date is populated, message is valid.
+            if (!goDate.isValid() && !stopDate.isValid()) {
```
ChristianMurphy: copying from https://github.com/uPortal-Project/uportal-app-framework/pull/641#discussion_r156768897

Moment comparisons return booleans so this can be flattened to

```js
var now = moment();
var goDate = moment(message.goLiveDate);
var stopDate = moment(message.expireDate);

// If both dates are defined, check between
if (goDate.isValid() && stopDate.isValid()) {
  return now.isBetween(goDate, stopDate)
}

// If goLive defined, check after
if (goDate.isValid()) {
  return now.isAfter(goDate);
}

// If expire defined, check before
if (stopDate.isValid()) {
  return now.isBefore(stopDate);
}

// Has neither date defined, so don't filter it
return true;
```

components/portal/messages/controllers.js
Doug-Reed: Semantic change made.

components/portal/messages/filters.js
apetro: `ClojureScript`. If we really want to be experimenting with pure functions, where this is going is writing some of our stuff in `ClojureScript`. And hey then maybe it would be radically more testable.

apetro: Collapsing at least some of these conditionals might be more literate anyway.  The logic seems to be ORing some ways that the message might pass the filter? So maybe say that, with ye olde `||` operator?

components/portal/messages/filters.js
apetro: I like the comments inline. They don't go far enough. It's not readily apparent to my sleep-deprived brain what this filter does without a close reading, especially outside of this Pull Request with its supporting pull request description. Needs some kind of comment that says "Filters away not-yet-gone-live and expired messages." or so.

components/portal/main/partials/footer.html
```diff
@@ -21,14 +21,14 @@
...
+      <span>© <p>{{date | date:'yyyy'}}</p>, Board of Regents of the <a href="http://www.wisconsin.edu/" target="_blank" rel="noopener noreferrer">University of Wisconsin System</a></span>
     </div>
     <div layout="row" layout-align="center center" class="links-separated">
         <span ng-repeat="link in portal.theme.footerLinks">
-          <a ng-href='{{link.url}}' target='{{link.target}}' rel='noopener noreferrer' ng-click="pushGAEvent('footer-link',link.title,link.url + ' ' + link.target)">{{link.title}}</a>
+          <a ng-href="{{link.url}}" target="{{link.target}}" rel="noopener noreferrer" ng-click="pushGAEvent('footer-link',link.title,link.url + ' ' + link.target)">{{link.title}}</a>
         </span>
         <span ng-repeat="link in FOOTER_URLS">
-          <a ng-href='{{link.url}}' target='{{link.target}}' rel='noopener noreferrer' ng-click="pushGAEvent('footer-link',link.title,link.url + ' ' + link.target)">{{link.title}}</a>
+          <a ng-href="{{link.url}}" target="{{link.target}}" rel="noopener noreferrer" ng-click="pushGAEvent('footer-link',link.title,link.url + ' ' + link.target)">{{link.title}}</a>
```
apetro: I don't get it. Why are unrelated-to-date-filtering footer changes in this changeset?

components/portal/messages/filters.js
ChristianMurphy: > `ClojureScript`. If we really want to be experimenting with pure functions, where this is going is writing some of our stuff in `ClojureScript`. And hey then maybe it would be radically more testable.

There are a number of pure functional languages that compile to JavaScript.
https://github.com/stoeffel/awesome-fp-js#functional-languages-that-compile-to-javascript
https://stateofjs.com/2017/flavors/results

I've heard good things about Elm and Reason, as well as ClojureScript.

PR #641, opened 2017-12-13
feat: filter messages by dates (if provided)

### LINKED ISSUE #609: Honor goLiveDate and expireDate for notification-flavored messages
While the docs say 

```
goLiveDate: (optional) Accepts a simple ISO date, including time (as pictured). This is used to restrict displaying a message to a certain day/time.
expireDate: (optional) Accepts a simple ISO date, including time (as pictured). This is used to stop displaying a message at a certain day/time.
```

what they don't disclose is that this only works for announcement-flavored-messages. Fix to universally filter all flavors of messages when these optional fields are set.

thevoiceofzeke:
Resolves #609 

Duplicates/replaces #639 

**In this PR:**
- Filter messages by `goLiveDate` and `expireDate` if dates are provided
- If no dates provided, show the messages
- Break up some logic for the sake of separation of concerns

----

### DIFF
--- a/components/portal/messages/filters.js
+++ b/components/portal/messages/filters.js
@@ -18,7 +18,7 @@
  */
 'use strict';
 
-define(['angular'], function(angular) {
+define(['angular', 'moment'], function(angular, moment) {
   return angular.module('portal.messages.filters', [])
     .filter('separateMessageTypes', ['$filter', function($filter) {
       return function(messages) {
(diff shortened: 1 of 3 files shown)


components/portal/messages/filters.js
```diff
@@ -103,6 +103,36 @@ define(['angular'], function(angular) {
         });
       };
     })
+    .filter('filterByDates', function() {
+      return function(messages) {
+        // Get current date/time
+        var now = moment();
+        // Filter messages based on date/time
+        return messages.filter(function(message) {
+          if (message.goLiveDate || message.expireDate) {
```
ChristianMurphy: :confused: I'm a bit confused, `Array.prototype.filter` expects a boolean value.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter
This returns `message`, which happens to be `Object` which happens to be truthy.
AFAICT all paths return `message` so nothing will be filtered out.

thevoiceofzeke: You're right about misusing `filter`. It would be more correct to just go with the good old `angular.forEach` in this case. 

However, not every path returns `message`. Cases that don't meet the criteria are not returned (thus not added to resulting array). For example: A notification that has both `goLiveDate` and `expireDate` specified, and `moment()` is not between those dates. If both dates are specified, it will stop at that block, not return `message`, and not proceed to the next conditional. If they're not both specified, it'll check first for `goLiveDate`. If that attribute is specified, we know there must not be an `expireDate` specified, so it's safe to only check whether the `goLiveDate` is in the past. If it's not specified, then we check for `expireDate`, at which point we know it's safe to only check against validity of that date.

The notification with id `sample-uportal-app-framework-expired-notification` is there to prove that this works, in lieu of a proper unit test.

ChristianMurphy: ahhh, sneaky
Moment comparisons return booleans so this can be flattened to

```js
// If both dates are defined, check between
if (message.goLiveDate && message.expireDate) {
  return now.isBetween(message.goLiveDate, message.expireDate)
}

// If goLive defined, check after
if (message.goLiveDate) {
  return now.isAfter(message.goLiveDate);
}

// If expire defined, check before
if (message.expireDate) {
  return now.isBefore(message.expireDate);
}

// Has neither date defined, so don't filter it
return true;
```

thevoiceofzeke:
Closing due to duplication

PR #679, opened 2018-01-25
Reflect page title in document title

### LINKED ISSUE #654: Meaningful page titles (WCAG2:2.4.2)
uPortal-app-framework should

1. Meet WCAG2:2.4.2 (meaningful page titles) for all pages in the framework (compliant ootb), and
2. Not prevent apps using uPortal-app-framework from meeting WCAG2:2.4.2 (meaningful page titles) for all paths in the specific app.

apetro:
Purpose:

* Set the document title to values describing the location in a `uPortal-app-framework` application as the user navigates within the application, thereby enabling improving usability generally and fulfilling `WCAG2: 2.4.2` specifically (Page Titled: Web pages have titles that describe topic or purpose. (Level A)).

Done:

* Adds title computation to `mainService`, implementing our (goofy?) "rule" about page naming, e.g. `PageName | AppName | PortalName`, with oodles of failing gracefully, and unit testing.
* Refactors title setting in `PortalMainController`, relying upon `mainService` to compute the title and only gathering the inputs and doing the view (Document) modification in the controller layer. This continues to mutate document title on theme change, since theme change can mean a change in portal name. Unit tests the initial document title setting.
* Adds title setting in `AppHeaderOptionsController`, with unit test.
* In `app-header` directive calls the title setting in `AppHeaderOptionsController` to set the document title reflecting the title in the app header, which in some contexts corresponds to what the user would regard as the "page title" (and failing that, at least reflects the name of the app).

app-header-sets-page-title

Deferred:

* Unit test the `PortalMainController` reacting to theme change by resetting document title. (This changeset didn't introduce this feature and hasn't so far as I can tell broken it.)
* Unit test `app-header` directive (didn't get to figuring out how to test directives and that directive is moribund.)
* Maybe do something less hacky than set the document title as the side effect of a div. (There's gotta be a more idiomatic way for a directive to call a controller function on rendering.)
* Reconcile with parallel efforts to eliminate the `app-header` directive ( #682 ). Will have to face down what that bigger change means for setting the Document title in the course of that bigger change.

----

Resolves #654 .

----

### DIFF
--- a/components/portal/main/controllers.js
+++ b/components/portal/main/controllers.js
@@ -39,7 +39,7 @@ define(['angular', 'require'], function(angular, require) {
      * sets the document title to that value.
      * @param {string} pageTitle - Optional, name of specific page viewed.
      */
-    function updateTitle(pageTitle) {
+    function updateWindowTitle(pageTitle) {
       var appTitle = NAMES.title;
 
       var portalTitle = '';
@@ -67,13 +67,13 @@ define(['angular', 'require'], function(angular, require) {
       $scope.APP_OPTIONS = APP_OPTIONS;
 
       if (NAMES.title) {
-        updateTitle();
+        updateWindowTitle();
       }
       // https://github.com/Gillespie59/eslint-plugin-angular/issues/231
       // eslint-disable-next-line angular/on-watch
       $rootScope.$watch('portal.theme', function(newValue, oldValue) {
         if (newValue && newValue !== oldValue) {
-          updateTitle();
+          updateWindowTitle();
         }
       });
 
--- a/components/portal/main/controllers.js
+++ b/components/portal/main/controllers.js
@@ -35,6 +35,9 @@ define(['angular', 'require'], function(angular, require) {
 
     /**
      * Set Document title.
+     * Asks mainService what the document title ought to be and
+     * sets the document title to that value.
+     * @param {string} pageTitle - Optional, name of specific page viewed.
      */
     function updateTitle(pageTitle) {
       var appTitle = NAMES.title;
--- a/components/portal/main/controllers.js
+++ b/components/portal/main/controllers.js
@@ -33,6 +33,26 @@ define(['angular', 'require'], function(angular, require) {
       layoutMode: 'list', // other option is 'widgets
     };
 
+    /**
+     * Set Document title.
+     */
+    function updateTitle(pageTitle) {
+      var appTitle = NAMES.title;
+
+      var portalTitle = '';
+      if ($rootScope.portal && $rootScope.portal.theme &&
+            $rootScope.portal.theme.title) {
+        // there's an active theme with a title.
+        // consider that title the name of the portal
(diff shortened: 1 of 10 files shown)


components/portal/main/services.js
```diff
@@ -64,20 +64,78 @@ define(['angular'], function(angular) {
...
+        if (appTitle !== portalTitle) {
+          // if the windowTitle already has content, first add a separator
+          if (windowTitle !== '') {
+            windowTitle = ' | ' + windowTitle;
+          }
+          windowTitle = appTitle + windowTitle;
+        }
+
+        // if there's a page name not redundant with the app name, prepend it.
+        if (pageTitle && pageTitle !== '' && pageTitle !== appTitle) {
```
davidmsibley: `if (pageTitle && pageTitle !== appTitle) {` Should be enough, check the `If()` tab here: https://dorey.github.io/JavaScript-Equality-Table/

components/portal/main/services.js
```diff
@@ -64,20 +64,78 @@ define(['angular'], function(angular) {
...
+      function setTitle(pageTitle) {
+        var windowTitle = ''; // we finally set the window title to this.
+
+        // first, gather pieces of the desired window title
+
+        var portalTitle = ''; // the name of the portal if we can determine it
+        var appTitle = ''; // the name of the app
+
+        if ($rootScope.portal && $rootScope.portal.theme &&
+              $rootScope.portal.theme.title) {
```
davidmsibley: if we want to do something with the theme title, we should pass it in via parameters. `$rootScope` or even `$scope` shouldn't bleed out to services

components/portal/main/controllers.js
```diff
@@ -33,6 +33,26 @@ define(['angular', 'require'], function(angular, require) {
       layoutMode: 'list', // other option is 'widgets
     };
 
+    /**
+     * Set Document title.
+     */
```
vertein: I see that we call updateTitle with no param later in this file.  Maybe a comment on what happens when no param is called?  What happens with the parameter?

components/portal/main/controllers.js
```diff
@@ -33,6 +33,29 @@ define(['angular', 'require'], function(angular, require) {
       layoutMode: 'list', // other option is 'widgets
     };
 
+    /**
+     * Set Document title.
+     * Asks mainService what the document title ought to be and
+     * sets the document title to that value.
+     * @param {string} pageTitle - Optional, name of specific page viewed.
```
ChristianMurphy: :art: jsDoc syntax for optional is
```javascript
/** @param {string} [pageTitle] - name of specific page viewed
```

reference: http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

apetro:
This framework change seems to have worked out well in `uPortal-home` with zero change in that product: most pages use `app-header` with a label describing the page and so most pages now have a sensible Document title.

title-the-pages

PR #683, opened 2018-02-01
feat: add analytics to widget interactions

davidmsibley:
We've got outbound link analytics, but we don't have any analytics on internal widget interactions. This adds them to `a` tags within expanded and compact widgets.

I broke with outbound link style and used the href as the event label. I'd welcome an opinion as to why hrefs should be the event action.

----

### DIFF
--- a/components/portal/widgets/directives.js
+++ b/components/portal/widgets/directives.js
@@ -24,7 +24,7 @@ define(['angular', 'require'], function(angular, require) {
   /**
    * Just the widget card -- gets the widget type from scope
    */
-  .directive('widget', function() {
+  .directive('widget', ['miscService', function(miscService) {
     return {
       restrict: 'E',
       transclude: true,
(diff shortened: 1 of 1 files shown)


components/portal/widgets/directives.js
```diff
@@ -33,13 +33,26 @@ define(['angular', 'require'], function(angular, require) {
       },
       templateUrl: require.toUrl('./partials/widget-card.html'),
       controller: 'WidgetCardController',
+      link: {
+        post: function(scope, element, attrs) {
+          element.on('click', function(event) {
+            var el = event.target;
+            while (el && (!el.tagName || el.tagName.toLowerCase() !== 'a')) {
```
ChristianMurphy: :question: Will there always be a parent link tag?
I'd be surprised if there aren't some areas that are using a button.

components/portal/widgets/directives.js
```diff
@@ -48,8 +61,22 @@ define(['angular', 'require'], function(angular, require) {
       },
       templateUrl: require.toUrl('./partials/compact-widget-card.html'),
       controller: 'WidgetCardController',
+      link: {
+        post: function(scope, element, attrs) {
```
ChristianMurphy: :question: Could analytics capture be provided as a service that could be applied to any directive?

components/portal/widgets/directives.js
davidmsibley: good catch. I just checked in widget-creator, almost everything is using an anchor, except for action items.

... that being said, I wonder if action items should actually have anchors instead.

apetro:
Historically I've used the label to try to differentiate different places in the UI a link was clicked from. Did this in [the ABE notifications analysis](https://docs.google.com/document/d/1BTOMCpVxhU4ZpH9XTxjxMMA5WyWnbGkQer5JJcSZnfg/edit), e.g. This solution is using more specificity than "Outbound Link" instead, capturing the fname and the widgetness and even the widget compact-or-expanded state. That's probably better than trying to reconstruct what labels were used where when.

PR #687, opened 2018-02-03
docs(changelog): copy edits

apetro:
+ prefer talking in terms of apps in third person rather than in terms of what "you" are using and "your" apps.
+ remove "Removed" section from pending unreleased changelog, since we're not using that section to document removing things. :smile:
+ wrap at 80 characters

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,16 +8,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Breaking changes
 
-+   Removed `<app-header>` directive. This affects any framework apps using that directive outside of `<frame-page>`. If you 
-are already using `<frame-page>`, you should get the new treatment without breakage. (#684)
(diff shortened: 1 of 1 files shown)


apetro:
Certainly writing in the third person (and especially with passive voice) can obfuscate rather than clarify.

I don't think I object to "you" generally. I objected to this usage specifically.

I don't think anthropomorphizing apps helps. So reading

> If you are already using `<frame-page>`, you should get

as "you the app", well, that's goofy.

And reading it as "you the developer", well, *you* aren't using a directive, the app that you play some role in maintaining uses it, and *it* should get...

So could write this as something like

> If the app you are maintaining is already using `<frame-page>`, it should get

I wouldn't object to that on "you" grounds, but I'd still prefer 

> `<frame-page>` usages should be unaffected.

for brevity.

Gotta tailor the documentation to the audience. This is BREAKING CHANGE documentation in a CHANGELOG. There's a premium on brevity, directness, and clarity on just what is and is not broken with the change. That's what I'm going for in that copy edit.

PR #698, opened 2018-02-13
WIP: fix: "Skip to main content" skip closer to main content

apetro:
Problem being solved: existing "Skip to main content" link doesn't skip enough of the header stuff, leaving the user with more stuff to step through before reaching the main content.

Solution:

+ Invents a new `skip-to-here` ID closer to what a user would regard as the beginning of the main content, in the `frame-page` directive.
+ Changes the "Skip to main content" link to link to that ID.

Locally tested using `npm run static:dev`, but the proof will be in seeing `uPortal-home` built on a `uPortal-app-framework` with this change.

WIP because I'm seeing a lot of header vertical space that I'd like to prove this didn't cause before offering this for merge.

----

PR considerations checklist:

### DIFF
--- a/components/portal/misc/partials/frame-page.html
+++ b/components/portal/misc/partials/frame-page.html
@@ -29,9 +29,12 @@
       </h1>
       <add-to-home ng-if="appShowAddToHome"></add-to-home>
     </div>
-    <div class="content-wrapper__push-content" ng-if="menuPushContent">
-      <ng-transclude></ng-transclude>
+    <div id="skip-to-here">
+      <div class="content-wrapper__push-content"
+        ng-if="menuPushContent">
+        <ng-transclude></ng-transclude>
+      </div>
+      <ng-transclude ng-if="!menuPushContent"></ng-transclude>
     </div>
-    <ng-transclude ng-if="!menuPushContent"></ng-transclude>
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


components/portal/misc/partials/frame-page.html
```diff
@@ -29,9 +29,13 @@ <h1 class="md-display-1" ng-if="pageTitle">
       </h1>
       <add-to-home ng-if="appShowAddToHome"></add-to-home>
     </div>
-    <div class="content-wrapper__push-content" ng-if="menuPushContent">
+
+    <span id="skip-to-here"></span>
```
ChristianMurphy: An interesting approach :thinking: 
I've usually seen `skip-to-here` applied directly to a `<main>` tag.
Progress is progress :+1:

apetro: Yup. And *not* applying it directly to `<main>` probably means we're not getting `<main>` quite right. Which might be a bigger lift to fix.

At this point the "Beta Settings"-ness of the Beta Settings page is reflected in the page title, so arguably it's redundant to also say that in the main content. Figure anyone following the skip-to-main content link already knows they're on the Beta Settings page or at least could have known that if they wanted to and is ready to skip to the actual page content.

At least that's the hypothesis. I'm not sure this change is really going to help, but the inadequacy of our skip-to-main-content link was flagged in a (manual) a11y testing, so trying to do something about it. Will change and re-test.

PR #708, opened 2018-03-02
feat: trap focus order within opened side navigation

thevoiceofzeke:
[MUMUP-3295](https://jira.doit.wisc.edu/jira/browse/MUMUP-3295): "As a person navigating MyUW via keyboard, I would like focus to remain in (cycle through) the open hamburger menu until I dismiss the menu, so that the menu is more usable and so that my navigation is less confusing."

**In this PR**:
- When normal side navigation menu (i.e. _not_ push-content version) is open, keyboard navigation is trapped within the menu
- Tabbing past the bottom-most item results in focus resuming at the top-most item, and vice versa

### Demo
sidenav-trap

----

PR considerations checklist:

### DIFF
--- a/components/portal/main/controllers.js
+++ b/components/portal/main/controllers.js
@@ -217,19 +217,27 @@ define(['angular', 'require'], function(angular, require) {
       vm.openMenuByDefault = false;
       vm.showMessagesFeatures = true;
 
+      /**
+       * Listen for unseen notifications
+       */
       $scope.$on('HAS_PRIORITY_NOTIFICATIONS', function(event, data) {
         if (angular.isDefined(data.hasNotifications)) {
           vm.hasPriorityNotifications = data.hasNotifications;
         }
       });
 
+      /**
+       * Listen for unseen announcements
+       */
       $scope.$on('HAS_UNSEEN_ANNOUNCEMENTS', function(event, data) {
         if (angular.isDefined(data.hasNotifications)) {
           vm.hasUnseenAnnouncements = data.hasAnnouncements;
         }
       });
 
-      // Close side nav on scroll to avoid awkward UI
+      /**
+       * Close side nav on scroll to avoid awkward interaction
+       */
       $window.onscroll = function() {
         if (vm.isMenuOpen() && !$mdMedia('gt-sm')) {
           vm.closeMainMenu();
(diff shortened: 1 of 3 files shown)


components/portal/main/controllers.js
```diff
@@ -253,6 +261,27 @@ define(['angular', 'require'], function(angular, require) {
         }
       };
 
+      /**
+       * Intercept focus when it would move outside the side nav,
+       * then reset it to the desired position within the side nav.
+       * @param {object} event
+       * @param {string} position
+       */
+      vm.trapFocus = function(event, position) {
```
apetro: Feasible to cover this with unit test?

thevoiceofzeke:
@ChristianMurphy The reason we can't used the tabindex solution is because we have no control of the tabindex of content include by the `ng-transclude` on the frame page (e.g. content from apps using the framework). Using `tabindex="-1"` on the containing element isn't enough, because focus will continue to look all the way down the tree. We'd still have to use Javascript to add `tabindex="-1"` to every DOM element outside the menu (and remove it from every element when the menu closes), which doesn't seem like a better solution to me :/.

ChristianMurphy:
I don't follow
If a section is focused and has tabindex -1 it should hold focus regardless of the surrounding page.

Even if the page semantics somehow breaking the default behavior and extra work arounds are needed.
Its better to use the `tabindex` semantic rather than creating a copy in JavaScript that now has to be tested and maintained.

thevoiceofzeke:
Do you have a reference or working example of what you're talking about? I might be misunderstanding you.

If I follow you, I should set `tabindex="-1"` on the "section" that I want to _retain_ focus. In this case, if the side navigation menu is open, an element within it (which we can specify) receives focus. Setting `tabindex="-1"` on any of the elements containing the focused element (e.g. the outermost element of the menu), or on the focused element itself, has no effect. Keyboard navigation still proceeds through the DOM as usual.

ChristianMurphy:
> Do you have another reference or working example of what you're talking about?

https://hackernoon.com/its-a-focus-trap-699a04d66fb5#52df
Re-reading my original comments I was unclear, it's not JavaScript events in general the concern me. It's attaching the event to elements on the edges, instead of attaching the event listener to the parent.
The manual-ness of adding the extra elements and correctly linking them leaves it at risk of being broken the next time the menu is edited.
Its the way around that is to detect anytime that the focus is outside the parent area an pull it back in.
https://github.com/vigetlabs/react-focus-trap/blob/master/src/focal-point.js#L104-L107

> If I follow you, I should set tabindex="-1" on the "section" that I want to retain focus.

Right it should be out of the normal focus ordering.

PR #721, opened 2018-03-15
Remove jQuery ui sortable dependency

thevoiceofzeke:
**In this PR**:
- Remove jQuery UI Sortable

----

PR considerations checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Versioning](http://semver.org/spec/v2.0.0.html).
   cases (#715)
 + Core Infrastructure Initiative badge (#711)
 
+### Changed
++ Replaced jQuery UI sortable with angular-drag-and-drop-lists
+
 ### Fixed
 
 + Fixed widget in-app messagge example (#714)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 * Core Infrastructure Initiative badge (#711)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ Versioning](http://semver.org/spec/v2.0.0.html).
   cases (#715)
 + Core Infrastructure Initiative badge (#711)
 
+### Changed
++ Replaced jQuery UI sortable with angular-drag-and-drop-lists
```
apetro: What the expected impact on applications developed using `uportal-app-framework`? Is this a breaking change? A no-op?

thevoiceofzeke: Some fact-finding will have to be done to figure out if any downstream apps are using the ui sortable dependency, but so far I've been unable to find one that does.

thevoiceofzeke:
@vertein & @ChristianMurphy, see: https://github.com/uPortal-Project/uportal-home/pull/795

CHANGELOG.md
vertein: I see now.  Can we remove this from UAF and have this only in home?

apetro: Semantic versioning isn't a value judgment, it's just a way of describing reality.

Sounds like the reality is we're yanking a dependency that downstream apps probably aren't using but hypothetically could be using. I'd favor acknowledging this as a breaking change, clearly articulating what we changed, and relying on downstream app maintainers to determine whether the breaking change breaks anything that affects particular apps.

vertein: Could we just remove this dependency and not add it into the framework?  Just add it to uPortal-home.  1 less thing to worry about

PR #735, opened 2018-04-03
fix: handle zero-links edge case in list-of-links

apetro:
`list-of-links` widgets can be configured with an array of zero links. This is especially plausible when sourcing the list of links from a dynamic remote feed (and indeed actually happens in representing HRS Portlet roles as a list-of-links).

Handle this case, degrading gracefully to present `list-of-links` widgets with zero internal links as `basic` widgets.

Before: blank widget face.

After (demo included in localhost example set):

list-of-no-links

----

PR considerations checklist:

### DIFF
--- a/components/portal/widgets/partials/type__list-of-links.html
+++ b/components/portal/widgets/partials/type__list-of-links.html
@@ -18,10 +18,24 @@
     under the License.
 
 -->
-<div class="widget-body list-of-links">
+<!-- ZERO LINKS: emulate a basic widget -->
+<div ng-show="config.links.length == 0">
+  <a tabindex="-1"
+    ng-href="{{ widget.url }}"
+    target="{{ widget.target ? widget.target : '_self' }}"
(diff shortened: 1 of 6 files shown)


components/portal/widgets/partials/type__list-of-links.html
```diff
@@ -18,10 +18,24 @@
...
+    ng-href="{{ widget.url }}"
+    target="{{ widget.target ? widget.target : '_self' }}"
+    rel="noopener noreferrer"
+    class="basic-widget">
+    <div class="widget-icon-container"
+      layout="column" layout-align="center center">
+      <widget-icon></widget-icon>
+    </div>
+  </a>
+</div>
```
apetro: It'd be nice to refactor this for less duplication between this edge case of the `list-of-links` type `partial` and the embedded `basic` widget implementation in the `widget-card` partial. Maybe extract a `basicWidget` `directive`?

PR #753, opened 2018-04-24
WIP fix: removes loginOnLoad

Doug-Reed:
{Substantive content goes here. Summarize the changeset and *why* the changeset.}

----

PR considerations checklist:

- [ ] Updates `CHANGELOG.md` to reflect this PR's change.
- [ ] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

### DIFF
--- a/components/portal/main.js
+++ b/components/portal/main.js
@@ -434,7 +434,7 @@ define([
         $sessionStorage.portal = $sessionStorage.portal || {};
         configureAppConfig();
 
-        if (APP_FLAGS.loginOnLoad && !lastLoginValid()) {
+        if (!lastLoginValid()) {
           $http.get(SERVICE_LOC.loginSilentURL).then(function(result) {
             if (APP_FLAGS.debug) {
               $log.info(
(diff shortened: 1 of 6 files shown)


components/portal/main.js
```diff
@@ -434,7 +434,7 @@ define([
         $sessionStorage.portal = $sessionStorage.portal || {};
         configureAppConfig();
 
-        if (APP_FLAGS.loginOnLoad && !lastLoginValid()) {
```
davidmsibley: Since loginOnLoad was almost always false, we should be able to just drop lines 437-459

components/portal/main/services.js
```diff
@@ -37,7 +37,6 @@ define(['angular'], function(angular) {
         userPromise = prom
           .then(function(result, status) { // success function
             var person = result.data.person;
-            if (APP_FLAGS.loginOnLoad) {
```
davidmsibley: Since loginOnLoad is almost always false, instead of removing this if condition, we can just drop lines 40-53

PR #760, opened 2018-05-15
fix: adjust z-indices to make widget actions available

thevoiceofzeke:
[MUMUP-3345](https://jira.doit.wisc.edu/jira/browse/MUMUP-3345): "Allow users to remove widgets that have an overlay"

**In this PR**:
- Widget remove button has higher priority than widget overlays
- Top-down adjustment of z-indices throughout app CSS (with LESS variables)
    - Dialogs/toasts always on top
    - Top bar higher than content
    - Menus higher than the menu focus trap
    - Widget actions higher than widget overlay
- Removed a couple unused/unnecessary z-index statements

----

PR considerations checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,16 +15,18 @@ Versioning](http://semver.org/spec/v2.0.0.html).
  
 ### Changed
 * "About" page now sources meaningful content from `aboutPageURL` (#755)
+* z-index values adjusted from highest (101) to lowest (51) stated values to play nicer with Angular Material (#760)
 
 ### Fixed
 
 * When priority notification title is truncated, provide full title as tooltip
   (#754)
 * Fix route to newly-added version info page (#758)
-
+* Widgets with overlays can be removed (#760)
+ 
 ### Removed
 
-* no loger used LoginOnLoad option removed (#753)
+* No logger used LoginOnLoad option removed (#753)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
  
 ### Removed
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -15,16 +15,18 @@ Versioning](http://semver.org/spec/v2.0.0.html).
  
 ### Changed
 * "About" page now sources meaningful content from `aboutPageURL` (#755)
+* z-index values adjusted from highest (101) to lowest (51) stated values to play nicer with Angular Material (#760)
```
apetro: How impactful is this change? Is this going to feel like a breaking change to upgrading framework apps? What will app developers typically need to do in updating apps to a framework version with this change?

CHANGELOG.md
```diff
@@ -15,16 +15,18 @@ Versioning](http://semver.org/spec/v2.0.0.html).
...
 * When priority notification title is truncated, provide full title as tooltip
   (#754)
 * Fix route to newly-added version info page (#758)
-
+* Widgets with overlays can be removed (#760)
+ 
 ### Removed
 
-* no loger used LoginOnLoad option removed (#753)
+* No logger used LoginOnLoad option removed (#753)
```
apetro: Better, but still not quite right. :)

vertein: longer*

thevoiceofzeke: Oops...

CHANGELOG.md
thevoiceofzeke: It shouldn't make much of an impact. I would guess most downstream devs would be writing custom z-indices to make content sit on top of our top bar or other high-index content, in which case they would be unaffected. The only case in which they might be affected is if they have z-indices for content sitting between the top bar and some other content (like the menu), which seems unlikely.

PR #761, opened 2018-05-15
feat: Add and use Roboto web font

thevoiceofzeke:
[MUMUP-3349](https://jira.doit.wisc.edu/jira/browse/MUMUP-3349): "As a uportal-app-framework developer, I'd like the Roboto web font preferred by default, so that my app can better adhere to Google Material guidelines."

**In this PR:**
- Made Roboto font available 
- Default to using Roboto
- Removed unused/inaccessible CSS

### Screenshot
screen shot 2018-05-15 at 11 29 59 am

----

PR considerations checklist:

### DIFF
--- a/components/css/buckyless/general.less
+++ b/components/css/buckyless/general.less
@@ -19,11 +19,6 @@
 
 /* General MyUW Styles */
 
-body {
-  overflow-x: hidden;
-  overflow-y: auto;
-}
-
 a {
   transition: @transition-color;
 
--- a/components/css/themes/common-variables.less
+++ b/components/css/themes/common-variables.less
@@ -70,8 +70,5 @@
 @transition-all: all 0.4s cubic-bezier(0.25, 0.8, 0.25, 1);
 @transition-color: color 0.4s cubic-bezier(0.25, 0.8, 0.25, 1);
 
-/* Custom mix-ins */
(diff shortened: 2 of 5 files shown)


components/css/themes/common-variables.less
```diff
@@ -70,8 +70,5 @@
 @transition-all: all 0.4s cubic-bezier(0.25, 0.8, 0.25, 1);
 @transition-color: color 0.4s cubic-bezier(0.25, 0.8, 0.25, 1);
 
-/* Custom mix-ins */
-.border (@color: @grayscale3, @size: 1px) {
```
davidmsibley: are we not using this?

components/css/buckyless/general.less
```diff
@@ -19,11 +19,6 @@
 
 /* General MyUW Styles */
 
-body {
-  overflow-x: hidden;
-  overflow-y: auto;
-}
```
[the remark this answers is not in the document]
thevoiceofzeke: Because the framework's CSS is wrapped in the `.myuw` class, which is on a div element, No body or html CSS in these files would ever be used.

components/css/themes/common-variables.less
thevoiceofzeke: No, we're not using that.

PR #755, opened 2018-05-04
feat: about page gets meaningful content

thevoiceofzeke:
[MUMUP-3312](https://jira.doit.wisc.edu/jira/browse/MUMUP-3312): "As a user following the About link in the sidebar menu, I'd like useful information answering whatever question I had that made me go to 'About', so that my questions are answered."

**In this PR:**
- uportal-app-framework version info and app-specific info moved to "Session info" page (linked to from "About" page)
- "About" page populated with text/links from a JSON source
- Updated docs to explain configuration of new feature
- Removed unused "demo widgets" page and path
- Removed frustrating, baffling, nonsensical remark-lint "list-item-indent" rule so that I can submit merge requests without unfixable false positives ruining the build

### Screenshots
screen shot 2018-05-04 at 11 15 08 am
screen shot 2018-05-04 at 11 42 20 am

----

PR considerations checklist:

- [x] Updates `CHANGELOG.md` to reflect this PR's change.
- [x] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

[Apereo CLA roster]: http://licensing.apereo.org/completed-clas
[Contributor License Agreements]: https://www.apereo.org/licensing/agreements

PR #765, opened 2018-05-30
fix: make new route backward compatible

thevoiceofzeke:
**In this PR**:
- Rolls back inadvertent breaking change (#755) to the about/route(s) module
    - Now returns only one route, so any apps configured to use that module won't break on startup
- Moves "session-info" page to a new module ("help")
    - Apps that want to use that page will need to add the module and route when upgrading (reflected in change log)

*Why it's necessary: Apps that are already configured to use the framework-provided "about" page do so by loading the portal/about/route module. #755 renamed the file and changed the route module to return two routes, which causes downstream apps to fail on startup when they can't find the (single) about route they expect.*

----

PR considerations checklist:


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Added
 
 * New "session info" page containing the former content of the "About" page (uportal-app-framework version info, app info JSON) (#755)
+  * To use: Add the `'portal/help/routes'` module in your main.js file (#765)
```
apetro: This should probably also have a bullet under `Unreleased`-->`Fixed`.

PR #787, opened 2018-07-11
fix: rm unused not-unique "widget-removal" div id

apetro:
The div will exist many times in the page, yielding [duplicate div ids forbidden by HTML spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) .

Fortunately, we're not using this div (it has no references in `uportal-app-framework` nor in `uPortal-home`). So drop the `id` attribute, resolving the uniqueness constraint violation and simplifying the code. Win-win.

----

PR considerations checklist:


thevoiceofzeke:
This will resolve the problem of no-widget-content in predev, but I have no idea why this didn't cause problems until now... 

confused marky mark

PR #794, opened 2018-07-16
fix: use contextual menu in compact widgets

thevoiceofzeke:
[MUMUP-3374](https://jira.doit.wisc.edu/jira/browse/MUMUP-3374): "As a compact mode user, I would like the widgets to not all say 'Remove from home', so that I know what my widgets are and can use them confidently."

**In this PR**:
- Compact widgets now use contextual menu (like expanded widgets)
- Added a margin to example page heading

### Screenshots
screen shot 2018-07-16 at 12 53 17 pm

screen shot 2018-07-16 at 12 53 26 pm

----

PR considerations checklist:

- [ ] Updates `CHANGELOG.md` to reflect this PR's change.
- [ ] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].

### DIFF
--- a/components/portal/main/partials/example-page.html
+++ b/components/portal/main/partials/example-page.html
@@ -28,7 +28,7 @@
       margin: 4px 8px;
     }
   </style>
-  <h2>Example expanded mode widgets</h2>
+  <h2 style="margin-left:26px">Example expanded mode widgets</h2>
   <div layout="row" layout-align="center start" style="flex-wrap:wrap;padding:18px;">
       <div flex-xs="100" class="widget-container">
         <widget fname="sample-widget__benefits"></widget>
--- a/components/portal/widgets/partials/compact-widget-card.html
+++ b/components/portal/widgets/partials/compact-widget-card.html
@@ -19,13 +19,24 @@
 
 -->
 <md-card class="list-content" id="widget-id-{{::widget.nodeId}}">
-  <md-button class="widget-action widget-info md-icon-button" aria-label="{{ widget.title }}: {{ widget.description }}" role="tooltip" ng-hide="widget.lifecycleState === 'MAINTENANCE'">
-    <md-tooltip md-direction="top" class="widget-action-tooltip">
-      {{ widget.description }}
-    </md-tooltip>
-    <md-icon>info</md-icon>
-  </md-button>
-  <div ng-transclude></div>
+  <!-- Widget contextual menu -->
+  <md-menu class="widget-action" md-position-mode="target-right bottom">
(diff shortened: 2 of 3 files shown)


components/portal/main/partials/example-page.html
```diff
@@ -28,7 +28,7 @@
       margin: 4px 8px;
     }
   </style>
-  <h2>Example expanded mode widgets</h2>
+  <h2 style="margin-left:26px">Example expanded mode widgets</h2>
```
apetro: Maybe do the same in line 58 for the "Example compact mode widgets" heading, for consistency? Maybe there should be a CSS class to abstract this style for re-use or so?

components/portal/widgets/partials/compact-widget-card.html
```diff
@@ -19,13 +19,24 @@
...
-    <md-icon>info</md-icon>
-  </md-button>
-  <div ng-transclude></div>
+  <!-- Widget contextual menu -->
+  <md-menu class="widget-action" md-position-mode="target-right bottom">
+    <md-button class="md-icon-button" aria-label="open {{ widget.title }} menu" ng-click="$mdOpenMenu($event)">
+      <md-tooltip class="widget-action-tooltip" md-direction="top" role="tooltip">
+        Open {{ widget.title }} menu
+      </md-tooltip>
+      <md-icon>more_vert</md-icon>
```
apetro: While this menu opener button *is* in the tab order when keyboard navigating, the icon does not visually appear when this button is selected via keyboard tab. It'd presumably be more usable if the button did appear when selected, so that users navigating a typical browser via keyboard have a stronger visual cue as to what they've selected. (The tooltip does appear, so it's kind of a tooltip-on-an-invisible-menu experience.)

image 2018-07-17 at 7 15 04 am

Maybe the feature would be more discoverable and usable in general if the menu actuator were to always show, rather than appearing on hover (or selection).

This changeset doesn't introduce this issue (it was already this way for expanded mode widgets), just continues it.

thevoiceofzeke: >Maybe the feature would be more discoverable and usable in general if the menu actuator were to always show, rather than appearing on hover (or selection).

Now that there's slightly less "stuff" in the heading space, I like this change. Added it with screenshot example.

components/portal/main/partials/example-page.html
thevoiceofzeke: Added to the second heading tag. I would argue for no CSS class because the presentation of the heading should be left up to whatever app is being developed with the framework (as such a class would *only* exist for the example page, for now).

PR #797, opened 2018-07-17
fix: make remove button focusable when present

thevoiceofzeke:
This one was fun!

**In this PR**:
- Use optional named slot for widget and compact widget `ng-transclude` 
- Update tabindex of transclude div to make it focusable when it has content
- Give transclude div an aria role so screen readers will read the button's aria label on keyboard focus

----

PR considerations checklist:


components/portal/widgets/controllers.js
```diff
@@ -94,6 +96,10 @@ define(['angular', 'moment'], function(angular, moment) {
             $scope.widget = data;
             $scope.widgetType = widgetType(data);
           }
+          // If there's a remove button, make it focusable by keyboard
+          if ($transclude.isSlotFilled('removeButton')) {
```
apetro: What if there's not a remove button, but there is other content in that menu? An about link? A widget-specific help link? Like, I'm not sure what all this menu might be used for in the future. Is this solution going to work if there's something interesting in this menu but that something isn't a remove button?

thevoiceofzeke: >Is this solution going to work if there's something interesting in this menu but that something isn't a remove button?

No, it won't work if there's something else. We could use an unnamed `transclude` as well to allow any random content, but since the contextual menu introduces a change in how that content is presented, I'd suggest avoiding that until there's a use case for it, since it would be difficult/impossible to ensure that such content would have a good UX for screen readers and keyboard users.

PR #798, opened 2018-07-17
MUMMNG-4226 add maintenance message

Doug-Reed:
Adds optional `widgetConfig.maintenanceMessage` that when set customizes the message shown in an expanded mode widget when the widget's backing `portlet-definition` is in `MAINTENANCE` lifecycle state.

Default unchanged:

default-maint-message

New custom message support:

custom-maint-message

Updated localhost examples to demonstrate both (and to demonstrate that this does not affect compact mode widget display):

maint-message-examples

----

PR considerations checklist:

- [X] This Contribution is under the terms of Individual [Contributor License Agreements][] (and also Corporate Contributor License Agreements to the extent applicable) appearing in the [Apereo CLA roster][].


components/portal/widgets/partials/widget-card.html
```diff
@@ -26,7 +26,8 @@
       <p class="overlay-warning-icon">
         <md-icon class="md-warn">warning</md-icon>
       </p>
-      <p>This app is undergoing maintenance and is temporarily unavailable.</p>
+      <ng-show="widget.maintenanceMessage"><p>{{ widget.maintenanceMessage }}</p></ng-show>
```
apetro: I don't think this is going to work. As I understand it, `ng-show` is idiomatically used as an attribute on the element that ought to be conditionally shown. As in,

```
<p ng-show="widget.maintenanceMessage">{{ widget.maintenanceMessage }}</p>
```

[Apparently it can be used as an element](https://docs.angularjs.org/api/ng/directive/ngShow) (was news to me), but that still needs both an element and the attribute. As in,

```
<ng-show ng-show="widget.maintenanceMessage"><p>{{ widget.maintenanceMessage }}</p></ng-show>
```

PR #820, opened 2018-08-29
feat(switch): add "switch" widget type

apetro:
Adds a new `switch` widget type, a meta widget that switches to become another widget type depending upon a dynamic value.

+ The `widgetURL` is the source for the dynamic JSON to evaluate.
+ Evaluates an AngularJS `expression` on that JSON to extract a value.
+ Matches that value against one or more `cases` configured in `widgetConfig`.
+ The matching case, or a default `caseotherwise` case, may specify a new `launchText`, widget type, `widgetConfig`, or even a new URL from which to source JSON to drive the newly selected widget type.

See [documentation for the new widget, added in this changeset](https://github.com/uPortal-Project/uportal-app-framework/blob/eccf6b0ea0a9a4613f20ac3fea0d1a7e40c94417/docs/make-a-widget.md#switch-widget-type).

Adds example widget publications in `localhost` to demonstrate this new widget type.

localhost-example-switch-widgets

Tech debt incurred:

+ None of this is covered with unit tests. We have a test coverage problem in this project.
+ ~Duplicates a method from `WidgetCardController` into the new `SwitchWidgetController` to support the way templating for `basic` type widgets determines the URL of the links in that widget.~ Not incurred by this changeset, but observation for future refactoring: Arguably this should be refactored to resolve that URL higher up in the architecture, rather than the controller being responsible to figure it out while rendering the widget. As in, shouldn't something higher up (`WidgetService`?) just go ahead and set `widget.url` to the right value and then the controller / partials / whatevers just rely on it?
+ ~Duplicates a chunk of the `widget-card` partial to re-implement the switching among widget types for content. Arguably this should be refactored to eliminate this duplication.~

Anticipated remaining refactoring in this PR:

- [x] Introduce `<basic-widget>` directive and accompanying controller and partial. Move `renderURL()` function to that controller, and thereby refactor away the duplication of that method introduced by this PR.
- [x] Refactor the switch widget partial to be a more general `widget-content` partial. (It's probably already that, literally just needs renamed, once `<basic-widget>` is a thing.) Reference it, probably via a directive, from `widget-card`, and thereby refactor away the duplication of that `ng-switch` introduced by this PR. 

----

PR considerations checklist:

### DIFF
--- a/components/portal/widgets/controllers.js
+++ b/components/portal/widgets/controllers.js
@@ -715,7 +715,7 @@ define(['angular', 'moment'], function(angular, moment) {
        * Fetch additional widget data
        */
       var initializeSwitchWidget = function() {
-      $log.warn("Initializing switch widget.");
+      $log.warn('Initializing switch widget.');
       // save widgetConfig as switchConfig so switch can then conditionally
       // mess with widgetConfig in activating a different widget type
       $scope.switchConfig = $scope.widget.widgetConfig;
@@ -729,7 +729,7 @@ define(['angular', 'moment'], function(angular, moment) {
         widgetService.getWidgetJson($scope.widget).then(function(data) {
           // if there's a case matching the returned JSON, activate it
 
-          $log.warn("Switch widget got JSON " + data);
+          $log.warn('Switch widget got JSON ' + data);
 
           var parsedExpression = $parse($scope.switchConfig.expression);
 
@@ -747,7 +747,7 @@ define(['angular', 'moment'], function(angular, moment) {
           if (!caseToActivate && $scope.switchConfig &&
             $scope.switchConfig.caseotherwise) {
             $log.debug($scope.widget.fname +
-              " switching to otherwise case after no case matched value " +
+              ' switching to otherwise case after no case matched value ' +
               dynamicValueToMatch);
             caseToActivate = $scope.switchConfig.caseotherwise;
           }
@@ -765,7 +765,7 @@ define(['angular', 'moment'], function(angular, moment) {
               // switch to the new type
               $scope.widget.widgetType = caseToActivate.widgetType;
             } else {
-              $scope.widget.widgetType = "basic";
+              $scope.widget.widgetType = 'basic';
             }
 
             if (caseToActivate.widgetConfig) {
@@ -785,35 +785,35 @@ define(['angular', 'moment'], function(angular, moment) {
           } else {
 
             $log.debug($scope.widget.fname +
-              " did not switch to any case (not even an otherwise case) " +
-              "for activation, so falling back to being a basic widget.");
+              ' did not switch to any case (not even an otherwise case) ' +
+              'for activation, so falling back to being a basic widget.');
 
-            $scope.widget.widgetType = "basic";
-            $scope.widget.widgetConfig = "";
+            $scope.widget.widgetType = 'basic';
+            $scope.widget.widgetConfig = '';
           }
 
         }).catch(function(error) {
 
           $log.error($scope.widget.fname +
-            " errored fetching or processing JSON [" + $scope.widget.widgetURL
-            + "] to switch on; falling back on being a basic widget.");
+            ' errored fetching or processing JSON [' + $scope.widget.widgetURL
+            + '] to switch on; falling back on being a basic widget.');
           $log.error(error);
 
-          $scope.widget.widgetType = "basic";
-          $scope.widget.widgetConfig = "";
+          $scope.widget.widgetType = 'basic';
+          $scope.widget.widgetConfig = '';
         }).finally(function() {
           // nothing to do
         });
       } else {
 
-        $log.warn($scope.widget.fname + " not configured with " +
-        "URL from which to load JSON [" + $scope.widget.widgetURL + "]," +
-        " expression [" + $scope.switchConfig.expression + "]" +
-        " to parse from that JSON, or cases to match against," +
-        " so falling back on being a basic widget.");
+        $log.warn($scope.widget.fname + ' not configured with ' +
+        'URL from which to load JSON [' + $scope.widget.widgetURL + '],' +
+        ' expression [' + $scope.switchConfig.expression + ']' +
+        ' to parse from that JSON, or cases to match against,' +
+        ' so falling back on being a basic widget.');
 
-        $scope.widget.widgetType = "basic";
-        $scope.widget.widgetConfig = "";
+        $scope.widget.widgetType = 'basic';
+        $scope.widget.widgetConfig = '';
       }
     };
 
--- a/components/portal/widgets/controllers.js
+++ b/components/portal/widgets/controllers.js
@@ -156,6 +156,32 @@ define(['angular', 'moment'], function(angular, moment) {
       $scope.initializeWidget($scope.fname);
   }])
 
+  /**
+   * Controller for un-typed (basic) widgets.
+   * Sets launch button url for these widgets.
+   */
+  .controller('BasicWidgetController', [
+    '$scope', '$log', '$localStorage', 'widgetService',
+    function($scope, $log, $localStorage, widgetService) {
+      /**
+       * Set the launch button url for static content,
+       * exclusive mode, and basic widgets
+       * @return {*}
+       */
+      $scope.renderUrl = function renderUrl() {
+        var widget = $scope.widget;
+        // Check for static content or exclusive mode portlets (rare)
+        if (widget.altMaxUrl == false) {
+          if (widget.staticContent != null) {
+            return 'static/' + widget.fname;
+          } else if (widget.renderOnWeb || $localStorage.webPortletRender) {
+            return 'exclusive/' + widget.fname;
+          }
+        }
+        return widget.url;
+      };
+    }])
+
     // External Message Controller
   .controller('WidgetExternalMessageController', [
     '$scope', '$log', 'widgetService',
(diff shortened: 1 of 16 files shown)


apetro:
Present use case for this is:

+ Some employees have a "H" `enrollmentFlag` some of the time. When they do the most important thing a "Benefit Information" widget has to say to them is that they have a personal enrollment opportunity.
+ Many employees have an "O" `enrollmentFlag` for around a month each year. When they do the most important thing a "Benefit Information" widget has to say to them is that they have N days left to participate in annual open benefits enrollment.
+ Most employees have mostly boring benefits most of the year, such that a basic hyperlink to the "Benefit Information" app is an appropriate "Benefit Information" widget experience.

This `switch` approach will allow continuing to use the `time-sensitive-content` widget type to handle the annual benefit enrollment aspect while swapping in custom content for the event-driven enrollment opportunity aspect.

Alternatives include

+ a `remote` widget type, where all the config is remoted on the other end of a JSON callback, rather than configured in the `portlet-definition`
+ fancier layout templating that would swap in any of N versions of the widget (and associated portlet) depending upon stuff

For the specific use cases MyUW intends to use switch for, the latter is going to add unwanted complexity (it's bad enough that we have two publications of "Benefit Information", I definitely don't want to deal with N publications for all the different permutations of campus affiliation and benefits status and...)

The `remote` idea is probably a pretty good one that would add a lot of flexibility. We already have steps towards it in sometimes dynamically sourcing `list-of-links`. However, in the details, we have an existing enrollmentFlag JSON service in production, so if all goes well, this approach allows *only* developing and configuring a `switch` widget switching on the response from that JSON and not requiring any other change. Allows deferring developing the server-side piece that would answer a call for dynamic widget configuration.

ChristianMurphy:
> fancier layout templating that would swap in any of N versions of the widget (and associated portlet) depending upon stuff

I'd lean toward something like this.

Another possibility could be leveraging UI router, each view would be a state/sub-state https://ui-router.github.io/guide/states
Then the description of a valid route/widget has a more widely recognized definition.
Which could work, but fancier layout templating still seems like a better long term solution.

apetro:
> My concern is the generic naming and documentation implies that it can and should be used as a general purpose routing system.
> That is misleading, this is designed to handle a very specific use case.

I don't agree. This is a generally applicable, re-usable, flexible widget type. This is designed to solve the general problem of composite widget types, and in doing so enable solving a specific use case that has arisen in MyUW. There's nothing benefits-enrollment or HR-specific or MyUW-specific about this widget type. By using an AngularJS `expression` to determine which switch case to activate, it's wildly flexible.

With this `type` available, one might:

+ Implement a better user experience in the "Manager Time and Approval" widget in MyUW by switching to a `custom` type widget when it turns out the employee has no time approval roles, rather than (status quo) abusing `list-of-links` to try to tell the employee they have no relevant roles.
+ Implement dynamic "app-under-maintenance" experiences in widgets by conditionally switching to a custom widget telling the user about the outage, as an alternative to the existing approaches around putting the app into `MAINTENANCE` lifecycle state, and potentially in doing this introduce the feature of exempting some operations users from the switching so they can test the widget even during a maintenance window when it's not expected to in general work.
+ Tee up switching on or off an interesting widget as part of application lifecycle management, or more generally make a widget conditional on the availability of the JSON it depends upon.

Who knows what opportunities we might find to solve problems with this widget type? It becomes another hammer in the toolchest.

ChristianMurphy:
> * Implement a better user experience in the "Manager Time and Approval" widget in MyUW by switching to a custom type widget when it turns out the employee has no time approval roles, rather than (status quo) abusing list-of-links to try to tell the employee they have no relevant roles.
> * Implement dynamic "app-under-maintenance" experiences in widgets by conditionally switching to a custom widget telling the user about the outage, as an alternative to the existing approaches around putting the app into MAINTENANCE lifecycle state, and potentially in doing this introduce the feature of exempting some operations users from the switching so they can test the widget even during a maintenance window when it's not expected to in general work.
> * Tee up switching on or off an interesting widget as part of application lifecycle management, or more generally make a widget conditional on the availability of the JSON it depends upon.

It could be done with a switch, or it could be handled through a router.

> Who knows what opportunities we might find to solve problems with this widget type? It becomes another hammer in the toolchest.

Angular already has solutions for the selection/routing problem, [angular-router](https://docs.angularjs.org/api/ngRoute/service/$route) and [ui-router](https://github.com/angular-ui/ui-router).
Instead of using an existing solution, yet another routing solution is being added.
This feels like a case of [not invented here](https://en.wikipedia.org/wiki/Not_invented_here).

-0.9

ChristianMurphy:
> This isn't necessarily a problem for the router, we're not keying off of url path

A url would be nice to have, but is not required by UI-Router.

> I'm not sure it'd be feasible to have every widget be a view.

I don't follow, are there high order components being used in the project?
If by view you mean page, ui-router supports scoped and nested routes https://ui-router.github.io/ng1/tutorial/hellogalaxy

> we're just trying to get a personalized rendering of a widget from static configuration.

It sounds like something that should be handled by an API more than the UI :man_shrugging:

apetro:
Maybe it helps to look at this from a perspective of architectural layers:

+ app directory aka portlet registry: what are the things in the portal? ( fname `benefit-information` is a thing.)
+ layout: which of those things ought to plop onto the user's homepage? ( fname `benefit-information` is a thing on this user's homepage).
+ widget: what should the user see when viewing a thing? ( here's what happens when the user renders `benefit-information`)

This feature is about switching up what's rendered in the box. Given that `benefit-information` is a thing, and given that it's a thing for this user, what should it look like for this user? That's why it's happening as a widget type, at the widget level of the architecture.

ChristianMurphy:
> Maybe it helps to look at this from a perspective of architectural layers:
> *    app directory aka portlet registry: what are the things in the portal? ( fname benefit-information is a thing.)
> *    layout: which of those things ought to plop onto the user's homepage? ( fname benefit-information is a thing on this user's homepage).
> *    widget: what should the user see when viewing a thing?

I completely agree with all of this.

> This feature is about switching up what's rendered in the box. Given that benefit-information is a thing, and given that it's a thing for this user, what should it look like for this user?

It is deciding what the user should see by changing which widget will appear in the page. That is a layout concern.

> That's why it's happening as a widget type, at the widget level of the architecture.

I'm not going to block this, but I disagree this is how it should be handled.
-0.9

components/portal/widgets/controllers.js
```diff
@@ -700,6 +706,121 @@ define(['angular', 'moment'], function(angular, moment) {
     }
   }])
 
+  // SWITCH widget type
+  .controller('SwitchWidgetController', [
+    '$scope', '$log', '$parse', 'widgetService',
+    function($scope, $log, $parse, widgetService) {
+      $scope.switching = true;
```
apetro: Will this reliably set before [the widget-content partial reads it](https://github.com/uPortal-Project/uportal-app-framework/pull/820/files#diff-893932addcc0222b82433c64708b8c68R49) ?

components/portal/widgets/partials/widget-card.html
```diff
@@ -71,62 +71,8 @@
...
-        <launch-button data-href="{{ renderUrl() }}"
-                       data-target="{{ widget.target ? widget.target : '_self' }}"
-                       data-rel="noopener noreferrer"
-                       data-button-text="{{ widget.widgetConfig.launchText ? widget.widgetConfig.launchText : 'Launch full app' }}"
-                       data-aria-label="{{ 'Launch ' + widget.title }}">
-        </launch-button>
-      </div>
-
-    </div>
+      <widget-content app="widget" config="widget.widgetConfig">
```
davidmsibley: do you use the config attribute? in the new partial everything gets it from `widget.widgetConfig`

components/staticFeeds/sample-widget__switch.json
```diff
@@ -0,0 +1,103 @@
...
+                  "title": "Dynamic value was O",
+                  "href": "",
+                  "icon": "fa-at",
+                  "target": "_blank"
+                }
+              ]
+            }
+          }
+        ],
+        "caseotherwise": {
```
davidmsibley: normally I'd call this the base case, but this is clear enough.

components/portal/widgets/controllers.js
davidmsibley: Yep. I could look it up, but even if angular did things completely backward, dom is slow, this would be initialized correctly by then.

components/portal/widgets/partials/widget-card.html
apetro: I don't use the `config` attribute, at least not intentionally. It's a copy-and-paste emulating something I didn't really understand in the existing calls to the type-specific directives. :) There's maybe some opportunity here to eliminate that, since it's reliably available under `widget` and `widget.widgetConfig` seems to be the preferred way of getting at widget config anyway.

components/staticFeeds/sample-widget__switch.json
apetro: Maybe `default` is the typical keyword for this? e.g. [what ng-switch does](https://github.com/uPortal-Project/uportal-app-framework/pull/820/files#diff-893932addcc0222b82433c64708b8c68R67).

Though [JSTL uses c:otherwise](https://www.tutorialspoint.com/jsp/jstl_core_choose_tag.htm), probably what came to mind when I named that.

So, rename to "defaultCase" ?

davidmsibley: nope it's fine as it is

apetro: Cool. Posted https://github.com/uPortal-Project/uportal-app-framework/pull/826 as a place to think about the name for this key. I'd be fine with that resulting in a better name, or resulting in confirming that `caseotherwise` is the name to run with.

PR #830, opened 2018-09-12
feat: adding inactivity check

davidmsibley:
MUMMNG-4274 set inactivity to 30 minutes. This PR keeps track of the network calls to the server to have an approximate idea of when the session will time out due to inactivity. A dialog will appear for the last minute of the session:
screen shot 2018-09-12 at 9 39 28 am

The user can reload the page within that minute to reup the session, or they will be redirected to `MISC_URLS.loginUrl` when the minute is up.

Fun note on why I pulled the partial into the javascript: The partial didn't get loaded until the first time that the modal was rendered, which meant the browser had to go get it from the server. This had the effect of auto-renewing the session when the first timeout was about to occur.

----

PR considerations checklist:


components/portal/timeout/controllers.js
```diff
@@ -54,13 +67,65 @@ define(['angular'], function(angular) {
...
+    }
+
     /**
      * Trigger a Session Expired dialog box
      */
     function triggerDialog() {
       $mdDialog.show({
-        templateUrl:
-          'portal/timeout/session-expired.html',
+        template:
```
apetro: Suggest leaving a comment here about why this needs to be an inline template rather than a template externalized to a URL, otherwise it'll be tempting to refactor out this template in the future.

components/portal/timeout/controllers.js
```diff
@@ -21,12 +21,24 @@
...
+    'portalShibbolethService', 'sessionInactivityService',
   function(
-    $sessionStorage, $log, $location, $timeout,
-    $mdDialog, $window, MISC_URLS, portalShibbolethService
+    $sessionStorage, $log, $timeout, $interval,
+    $mdDialog, $window, APP_OPTIONS, MISC_URLS,
+    portalShibbolethService, sessionInactivityService
   ) {
+    var checkInactiveDelay = 1; // Minutes
+    var appInactiveTimeout = APP_OPTIONS.inactivityTimeout; // Minutes
```
apetro: Handle case where `APP_OPTIONS.inactivityTimeout` is not set?

apetro: Consider documenting `APP_OPTIONS.inactivityTimeout` in [configuration.md](https://github.com/uPortal-Project/uportal-app-framework/blob/master/docs/configuration.md#app_options)

components/portal/timeout/controllers.js
```diff
@@ -54,13 +70,69 @@ define(['angular'], function(angular) {
...
+
     /**
      * Trigger a Session Expired dialog box
      */
     function triggerDialog() {
       $mdDialog.show({
-        templateUrl:
-          'portal/timeout/session-expired.html',
+        /*
+         * this template is inline to avoid accidentally
```
apetro: 👍

docs/configuration.md
```diff
@@ -57,6 +55,10 @@ containing element for app view(s). A menu that doesn't use the frame-page
 directive yields a hamburger button that does nothing when clicked. See
 [frame-page.html][] to understand how
 this is constructed.
++ **inactivityTimeout**: the length in minutes a login session can remain
```
apetro: 👍

docs/configuration.md
```diff
@@ -57,6 +55,10 @@ containing element for app view(s). A menu that doesn't use the frame-page
...
+inactive before the server expires it. A dialog will show during the last
+minute of a session, prompting user action. If no action is taken, the user
```
apetro: I got curious about best practices about this, specifically whether 1 minute of warning is long enough. Looks like [Harvard's accessibility guidance on this point](https://accessibility.huit.harvard.edu/give-users-control-over-timeouts) uses 1 minute of warning in its example, seems like a good sign that 1 minute is reasonable.

components/portal/timeout/controllers.js
```diff
@@ -54,13 +70,69 @@ define(['angular'], function(angular) {
...
+          '      <h2>Are you still there?</h2>' +
+          '    </div>' +
+          '  </md-toolbar>' +
+          '  <md-dialog-content class="md-dialog-content">' +
+          '    <p>Your session will expire soon. Please reload the ' +
+          'page to continue, otherwise you will be logged out.</p>' +
+          '  </md-dialog-content>' +
+          '  <md-dialog-actions layout="row">' +
+          '    <md-button aria-label="" ng-click="reloadPage()">' +
+          '      Reload' +
```
apetro: How feasible might it be to give the user a less invasive action to take than reloading the page outright? An async request that just tickles the server to extend the session timeout? [ssa.gov's guidance on timeout dialogs](https://www.ssa.gov/accessibility/bpl/bps/forms/timeout/default.htm), for instance, talks about advantages in returning the user back to what they were focused on as best one can when the user chooses to extend the session. Reloading the page is potentially going to discard whatever they were in the middle of working on behind this pop-up.

apetro: Cool. Better is better. Seems to me that this changeset already implements an experience better than the status quo even from a narrow perspective of accessibility. I'd therefore be supportive of a path forward that merges this increment of progress and backlogs the further improvement to keep users where they're at in the application in a re-upped session, if you'd want to go that route.

davidmsibley: yeah, I'm going to backlog an improvement for this.

PR #836, opened 2018-09-14
fix(actionitems): handle empty string quantities

apetro:
Enhances `action-items` to detect and handle the case where the JSON callback reports a `quantity` of the empty string.

After:

Screenshot of action-items widgets detecting and reporting as an error the empty string case

Before:

Screenshot of action-items widgets not detecting the empty string case

(Builds on #835 )

----

PR considerations checklist:


components/portal/widgets/controllers.js
```diff
@@ -376,9 +376,9 @@ define(['angular', 'moment'], function(angular, moment) {
           // Make sure we got a something back
           if (data) {
             // Make sure quantity is a number
-            if (isNaN(data.quantity)) {
```
apetro: Fun fact: [isNaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN)("") --> `true`.

thevoiceofzeke: 🤦‍♂️

PR #837, opened 2018-09-18
fix: use logoutURL for inactive timeout

davidmsibley:
Because the uPortal session hasn't actually died after an inactive timeout occurs, going directly to loginURL causes you to be automatically re-logged in. What we actually want is for the user to be explicitly logged out due to inactivity.

----

PR considerations checklist:

### DIFF
--- a/components/portal/timeout/controllers.js
+++ b/components/portal/timeout/controllers.js
@@ -103,7 +103,7 @@ define(['angular'], function(angular) {
      * Redirect to end the session;
      */
     function redirect() {
-      $window.location.replace(MISC_URLS.loginURL);
+      $window.location.replace(MISC_URLS.logoutURL);
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Added
 
-+ Added session inactivity checker, and updated logout dialog (#830)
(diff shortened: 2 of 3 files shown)


components/portal/timeout/controllers.js
```diff
@@ -103,7 +103,7 @@ define(['angular'], function(angular) {
      * Redirect to end the session;
      */
     function redirect() {
-      $window.location.replace(MISC_URLS.loginURL);
```
apetro: Handle case where logoutURL not configured, or figure it's always configured and so can be relied upon?

components/portal/timeout/controllers.js
```diff
@@ -103,7 +103,7 @@ define(['angular'], function(angular) {
...
+      $window.location.replace(MISC_URLS.logoutURL);
```
apetro: Handle case where logoutURL not configured, or figure it's always configured and so can be relied upon?

Thinking about this slightly longer, the thing to do might be to predicate the your-session-will-end-soon dialog feature on `MISC_URLS.logoutURL` being set, since if it's not set the front end can't do much reasonable on session timeout.

davidmsibley: It seems the documentation was the only thing requiring that, so thanks for reminding me to update the documentation.

I'm 50/50 on checking the logoutURL, if we do I'm thinking we do it in `portalShibbolethService.shibServiceActivated()` As of right now, it's just treated as a misconfiguration.

PR #864, opened 2018-10-30
fix: use less disruptive actions for inactivity timeout

davidmsibley:
On warning, prompt the user to make an ajax call, and restart the timer.
screen shot 2018-10-30 at 1 17 10 pm

On timeout, prompt the user to reload the page.
screen shot 2018-10-30 at 1 17 42 pm

----

PR considerations checklist:

### DIFF
--- a/components/portal/timeout/controllers.js
+++ b/components/portal/timeout/controllers.js
@@ -156,16 +156,16 @@ define(['angular'], function(angular) {
           };
           $scope.continueLabel = function() {
             return (isContinue)?'OK':'Reload';
-          }
+          };
           $scope.dialogText = function() {
             var result = 'Your session will expire soon due to inactivity. ' +
               'Please press OK to continue, otherwise you will be logged out.';
             if (!isContinue) {
               result = 'Your session has expired.' +
-                ' Please reload the page to continue.'
+                ' Please reload the page to continue.';
             }
             return result;
-          }
+          };
         },
       }).catch($log.error);
     }
--- a/components/portal/timeout/controllers.js
+++ b/components/portal/timeout/controllers.js
@@ -31,6 +31,8 @@ define(['angular'], function(angular) {
   ) {
     var checkInactiveDelay = 1; // Minutes
     var appInactiveTimeout = APP_OPTIONS.inactivityTimeout; // Minutes
+    var intervalRef;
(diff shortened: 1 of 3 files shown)


components/portal/timeout/controllers.js
```diff
@@ -138,15 +146,27 @@ define(['angular'], function(angular) {
...
             $mdDialog.hide();
           };
-          $scope.reloadPage = function() {
-            reload();
+          $scope.continueLabel = function() {
+            return (isContinue)?'OK':'Reload';
+          };
+          $scope.dialogText = function() {
+            var result = 'Your session will expire soon due to inactivity. ' +
+              'Please press OK to continue, otherwise you will be logged out.';
```
apetro: Microcopy: Given that we're not going to log people out of the IdP, I wonder if text more parallel with the after-expiration text would better set user expectations, something like 'Press OK to continue, otherwise your session will expire.'  , Then it's telling them the session will expire, and then a minute later, it's, hey, the session expired.

components/portal/timeout/controllers.js
```diff
@@ -31,6 +31,8 @@ define(['angular'], function(angular) {
   ) {
     var checkInactiveDelay = 1; // Minutes
     var appInactiveTimeout = APP_OPTIONS.inactivityTimeout; // Minutes
+    var intervalRef;
+    var isContinue = true;
```
apetro: Clever. I wouldn't have implemented this this way, but seeing it implemented this way, this adds some elegance.

apetro:
👍 . I like this a lot. It's a better experience and it's responsive to the feedback we received and to the policy/practice clarification we received, both quite recently, so it's in that vein of MyUW agility, responding to feedback.

docs/configuration.md
```diff
@@ -58,7 +58,7 @@ this is constructed.
 + **inactivityTimeout**: the length in minutes a login session can remain
 inactive before the server expires it. A dialog will show during the last
 minute of a session, prompting user action. If no action is taken, the user
-will be redirected to `MISC_URLS.logoutURL` when the session has expired.
+will be prompted to reload the page when the session has expired.
```
apetro: :+1: on keeping the docs in sync with the reality of what these settings do.

PR #890, opened 2019-03-22
docs: deprecate "priority" notifications

apetro:
Updates documentation to reflect that the "priority" notification feature is deprecated, getting slightly in front of intent to replace with a new banner message feature hewing a bit closer to (spirit of?) Material Design and reducing complexity of notifications as such.

----

PR considerations checklist:

### DIFF
--- a/components/css/buckyless/notifications.less
+++ b/components/css/buckyless/notifications.less
@@ -56,7 +56,7 @@
         .md-list-item-text {
           padding: 16px 0;
 
-          /* Priority notifications are deprecated. */
+          /* Priority notifications are DEPRECATED. */
           .notification__priority-message {
             color: @grayscale8;
             font-size: 12px;
--- a/components/css/buckyless/notifications.less
+++ b/components/css/buckyless/notifications.less
@@ -55,7 +55,7 @@
 
         .md-list-item-text {
           padding: 16px 0;
-
+          /* Priority notifications are deprecated. */
           .notification__priority-message {
             color: @grayscale8;
             font-size: 12px;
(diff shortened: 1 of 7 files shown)


components/css/buckyless/notifications.less
```diff
@@ -56,6 +56,7 @@
         .md-list-item-text {
           padding: 16px 0;
 
+          /* Priority notifications are deprecated. */
```
kedarjoyner: It looks like these deprecation comments in other files are uppercase. Should this be as well?

docs/messaging.md
```diff
@@ -46,7 +46,7 @@ notifications count and a link to the notifications page.
 On the notifications page, users can view, follow calls to action on, and
 dismiss their notifications. They can also click the "Dismissed" tab to view
 notifications they've previously dismissed. High priority notifications
-float to the top of the lists.
+(DEPRECATED) float to the top of the lists.
```
kedarjoyner: Expected to see the deprecated message at the end of the bulleted list item like the header of this section: 
`- float to the top of the lists (DEPRECATED)`

Could go either way, but would do one or the other just for consistency's sake. Messaging is very clear, and I would understand this as a new developer coming into the project.

components/css/buckyless/notifications.less
apetro: Good call on consistency. Normalized to "DEPRECATED".

docs/messaging.md
apetro: That's not a bulleted list item, though GitHub's diff treatment does make it look like a Markdown list item.

It's a sentence, changing 

> High priority notifications float to the top of the lists.

to 

> High priority notifications (DEPRECATED) float to the top of the lists.

So the DEPRECATED is placed close to the aspect that's deprecated.

I figure this is probably good enough for now, this may be short lived docs anyway if @thevoiceofzeke 's adding the banner message UI that conceptually replaces "priority" notifications continues to go well. :)

PR #909, opened 2019-05-30
feat: invent `list-of-links` `widgetConfig.suppressLaunchButton`

apetro:
There's long been a documented feature of `list-of-links` that would suppress the expanded mode widget launch button (bar) in the case where `widgetConfig.launchText` is not set.

That feature didn't actually work. But the Thumper Principle is a pretty good one. When the launch bar adds no value beyond the foregoing list-of-links, we should not show it. So this pull request (re-)adds the feature, this time triggering it more explicitly so that omitting `launchText` can continue to yield the default show-launch-button-with-default-text behavior.

----

PR considerations checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ and this project adheres to
 
 ## [unreleased][]
 
++ feature: in `list-of-links` widget type, new `widgetConfig` option
+  `suppressLaunchButton` which will make the widget omit the launch button
+  across the bottom of the widget.
++ documentation: stop documenting that omitting `launchText` in `list-of-links`
+  widgets would supress the widget launch button. This didn't work and instead
+  such list-of-links widgets would use default launch button text.
+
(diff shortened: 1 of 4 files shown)


apetro:
[Looks like](https://travis-ci.org/uPortal-Project/uportal-app-framework/jobs/539193028#L681) `commitlint` is objecting to `style: 2 space indent` because `subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]`. Which is odd, because `2` is neither sentence-case, start-case, pascal-case, nor upper-case. Maybe what it means to be objecting to is that subject starts with an integer. Which sounds like a bug, the commit message would be no more clear if it were `style: two space indent`.

Anyway, I don't think the continuous-integration/travis-ci/pr unsuccessful represents a real problem with this changeset, and the objectionable commit message will be `squash`ed away from `master` anyway.

PR #938, opened 2019-12-02
Decouple launch-button element from AngularJS framework

paulerickson:
This change decouples the 'launch-button' element/component from AngularJS by reimplementing it as a custom element (see [IAMYUW-66](https://jira.doit.wisc.edu/jira/browse/IAMYUW-66)).

I like to distinguish between custom elements that have a shadow DOM root as "web components" and those that don't as "custom elements".

I made this a custom element because the CSS affecting it is not currently componentized by AngularJS, so extracting and encapsulating its styles would be an orthogonal change (and also a pain in the ass).

Since the point is to decouple from AngularJS, I think this change is legit and can be used as an example, but I want to note here that I'm not sure componentization is even warranted in the first place.  This one element only encapsulates one element, adding virtually no behavior to it (viz. fixing the `rel` attribute and defaulting the `aria-label` attribute).

----

 - I ran `npm audit fix`, but there are still issues that require manual or breaking upgrades.

PR considerations checklist:


package.json
```diff
@@ -167,10 +169,12 @@
...
+      "padded-blocks": "off",
       "prefer-rest-params": "off",
       "prefer-spread": "off",
-      "valid-jsdoc": "warn",
+      "require-jsdoc": "off",
       "rest-spread-spacing": "off",
+      "valid-jsdoc": "warn",
       "yield-star-spacing": "off"
     }
   },
```
paulerickson: I'm suggesting some linting rule changes because
* ECMAScript 2015 is super-supported and still a very conservative bar.  The `class` keyword was introduced in 2015.  `async`/`await` is not introduced until a later version, and using it can lead to issues with AngularJS rendering.
* if we want to get away from angular, then we should reject rules that say "prefer angular's thing to the standard thing"
* padding blocks can by very readable when done thoughtfully (my rule of thumb is to pad the first two levels of blocks) and makes it possible to manipulate codeblocks as paragraphs in text editors like vim and emacs.
* jsdoc is valuable on APIs, but for internally-used code—esp. implementations of standard methods like `attributeChangedCallback` or property accessors—can be duplicative, desensitize the reader, and even mislead, when they are not updated.

apsummers:
Is this not the same thing as `myuw-card-footer` in [myuw-card](https://github.com/myuw-web-components/myuw-card)?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@uportal/app-framework",
-  "version": "12.2.0",
+  "version": "12.3.0",
```
apetro: Eh. I'm not sure about this bumping the `version` in the PR that adds a feature rather than in the process of cutting a release. This is going to make [the instructions for cutting the release](https://github.com/uPortal-Project/uportal-app-framework/blob/master/docs/releasing.md#node-release) not quite right, since those instructions involve running `npm version` to bump the version number, isn't it?

```suggestion
  "version": "12.2.0",
```

pom.xml
```diff
@@ -30,7 +30,7 @@
   <groupId>org.apereo.uportal</groupId>
   <artifactId>uportal-app-framework</artifactId>
   <packaging>war</packaging>
-  <version>12.2.1-SNAPSHOT</version>
+  <version>12.3.0</version>
```
apetro: Ditto comment above that bumping version number in the feature PR is not current process and is going to make [status quo documentation](https://github.com/uPortal-Project/uportal-app-framework/blob/master/docs/releasing.md#java-release) not quite applicable and is going to break status quo automation.

Leave this the SNAPSHOT and when this merges Jenkins will build and deploy a new SNAPSHOT which will trigger the chain that results in a uPortal-home based on this snapshot deployed as predev.

```suggestion
  <version>12.2.1-SNAPSHOT</version>
```

package.json
paulerickson: Thanks, I had not read those instructions.  I changed them back to 12.2.0 for the JS artifact and 12.2.1-SNAPSHOT for the Java artifact.

paulerickson:
> Is this not the same thing as `myuw-card-footer` in [myuw-card](https://github.com/myuw-web-components/myuw-card)?

@apsummers yeah, looks like the same part of the card.  However, this project kicks out a bunch of themed stylesheets that wouldn't (I presume) apply to its shadow DOM as expected.  _Also_, that component has no `observedAttributes`, so it doesn't set those values consistently.  When I tested by adding markup to a page, `href` was set, but not `text` (maybe `href` is handled by the superclass, or maybe its just a race condition), but what tends to happen in an AngularJS app is that some placeholder expression gets read initially, but then angular comes by and transcludes something else after construction, which in this case won't invoke the `attributeChangedCallback`.  Anyway, I can make a PR for that.

apetro:
Too stale. Closing un-merged. This theme of migrating from AngularJS Directives to Web Components is a good one, an active one, and something like this changeset will eventually happen, but probably not as implemented here.

PR #963, opened 2020-04-08
Upgrade myuw-search Web Component to version 1.5.1

nogalpaulina:


jimhelwig:
So the Travis CI build failed. I'm looking at this for the first time, but I think it is because the commit comments (or PR comments) don't follow the Conventional Commits standard. Are we intending to follow that standard or are we going to turn off that check?

apetro:
https://github.com/uPortal-Project/uportal-app-framework/pull/958 would drop Conventional Commits. I think in practice Conventional Commits isn't worth the hassle and we should merge #958 .

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to
 
 + No changes yet
 
+## 14.0.2 - 2020-04-08
```
apetro, line 13: [The most recent release was 14.0.0 released on April 1](https://repo1.maven.org/maven2/org/apereo/uportal/uportal-app-framework/), so the next version might be 14.0.1, and the 14.0.1 entry below isn't quite right, needing merged downward into 14.0.0 with a corrected release date.

PR #967, opened 2020-06-18
Myuw profile revision 2

nogalpaulina:
## 15.0.3

+ Update import path to use latest unpkg version
----

PR considerations checklist:

### DIFF
--- a/components/head-static.html
+++ b/components/head-static.html
@@ -48,8 +48,8 @@
 <script type="module" src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-search@1.5.1/myuw-search.min.mjs"></script>
 <script nomodule src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-search@1.5.1/myuw-search.min.js"></script>
 
-<script type="module" src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-profile@1.6.3/myuw-profile.min.mjs"></script>
-<script nomodule src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-profile@1.6.3/myuw-profile.min.js"></script>
+<script type="module" src="https://unpkg.com/@myuw-web-components/myuw-profile@latest?module"></script>
+<script nomodule src="https://unpkg.com/@myuw-web-components/myuw-profile@latest"></script>
 
 <link href="my-app/my-app.css" rel="stylesheet" type="text/css" />
 <link rel="shortcut icon" href="img/favicon.ico" type="image/x-icon" />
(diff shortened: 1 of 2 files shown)


components/head-static.html
```diff
@@ -48,8 +48,8 @@
 <script type="module" src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-search@1.5.1/myuw-search.min.mjs"></script>
 <script nomodule src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-search@1.5.1/myuw-search.min.js"></script>
 
-<script type="module" src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-profile@1.6.3/myuw-profile.min.mjs"></script>
-<script nomodule src="https://cdn.my.wisc.edu/@myuw-web-components/myuw-profile@1.6.3/myuw-profile.min.js"></script>
+<script type="module" src="https://unpkg.com/@myuw-web-components/myuw-profile@latest?module"></script>
```
apetro, line 51: There are tradeoffs in using the `@latest` alias. I'd like to be sure we're navigating the tradeoffs intentionally.

The huge advantage of `@latest` is that it will, well, automatically use the latest. So if we use this `<script` import in production, production will automatically use the latest `myuw-profile` release published to NPM. Won't have to cut an app framework release, won't have to cut a uPortal-home release, won't have to migrate anything, to pick up a new `myuw-profile` version. That's very powerful, and convenient, and maybe drives down release engineering effort.

The major disadvantage of using `@latest` is that it will, automatically, use the latest `myuw-profile` release. Not the latest release in the 1.x major version line that we've semantically versioned to be backwards compatible. The latest period. The act of publishing a later `myuw-profile` web component release to NPM becomes the act of adopting that release in production.

Traditionally, we've had rather lightweight change management on software product releases (mostly, if a developer thinks it's time to release, great, cut a release and let's get that up in test and qa and see what happens) and had heavier change management on migrating those software product releases to production. Traditionally, it's been really okay to cut any Web Component release or app framework release or uPortal-home release or Preparing for SOAR release that might plausibly work, because best case it will and we migrate it, and worst case we have a named thing to try out in test and QA and can concretely talk about what needs to change from what was in some specific version. It's all good.

I think the way to think about it is this: there has to be some amount of rigor somewhere in the system. If we have all the rigor we think we need in the migration to production, then we can afford to be loose (and, thus, faster and more scrappy) in development and release. If we've eliminated the role of migration in adding rigor, then whatever rigor we need, we'll have to have it in the development and release processes.

apetro, line 51:
```suggestion
<script type="module" src="https://unpkg.com/@myuw-web-components/myuw-profile@1.6.3/dist/myuw-profile.min.mjs"></script>
```

components/head-static.html
```diff
@@ -48,8 +48,8 @@
...
+<script nomodule src="https://unpkg.com/@myuw-web-components/myuw-profile@latest"></script>
```
apetro, line 52:
```suggestion
<script nomodule src="https://unpkg.com/@myuw-web-components/myuw-profile@1.6.3/dist/myuw-profile.min.js"></script>
```

PR #971, opened 2020-07-13
bugfix: before login, now not in guestMode (so the login button is sh…

cksteven:
…own)

{Substantive content goes here. Summarize the changeset and *why* the changeset.}

---
I found that this fix doesn't apply to the latest master branch, so this is unlikely to be merged.


cksteven:
outdated (resolved by another PR)

PR #979, opened 2020-08-17
bugfix: empty notifications page

cksteven:
- Update `myuw-notifications` to v1.3.3
- Fix page-breaking injection error
- As`md-list-item` appends empty `md-secondary` element to DOM, change style of `notifications-list-item`: `width: 100%`

### DIFF
--- a/components/portal/messages/controllers.js
+++ b/components/portal/messages/controllers.js
@@ -168,7 +168,7 @@ define(['angular'], function(angular) {
     ])
 
     .controller('NotificationsController', ['$q', '$log', '$document',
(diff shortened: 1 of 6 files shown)


components/portal/messages/controllers.js
```diff
@@ -168,7 +168,7 @@ define(['angular'], function(angular) {
     ])
 
     .controller('NotificationsController', ['$q', '$log', '$document',
-    '$scope', '$window', '$localStorage', '$filter', 'MESSAGES',
+    '$scope', '$window', '$rootScope', '$filter', 'MESSAGES',
```
cksteven, line 171: This mismatch was causing the page-breaking bug on predev

PR #980, opened 2020-08-19
Fix 'See all' notifications page styles

cksteven:
Fix 'See all' notifications page styles
- add a bottom border for each notification item
- add vertical padding for each notification item
- realign action buttons
- set high priority text bold

### DIFF
--- a/components/portal/messages/partials/view__notifications.html
+++ b/components/portal/messages/partials/view__notifications.html
@@ -48,7 +48,7 @@
                 <!-- DISMISS BUTTON -->
                 <div class="notifications-list__action">
                   <md-button class="md-icon-button"
-                             ng-hide="notification.dismissible === false"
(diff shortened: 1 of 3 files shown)


components/portal/messages/partials/view__notifications.html
```diff
@@ -48,7 +48,7 @@
                 <!-- DISMISS BUTTON -->
                 <div class="notifications-list__action">
                   <md-button class="md-icon-button"
-                             ng-hide="notification.dismissible === false"
+                             ng-style="{'visibility' : notification.dismissible === false  ? 'hidden' : 'visible'}"
```
nogalpaulina, line 51: I don't think that we need to make this change. I would keep ng-hide.

cksteven: This change will fix the alignment of action buttons. I could try other ways to fix it, but this seems easiest... What do you think?

PR #997, opened 2020-10-29
Test custom error message on Student Orientation widget

nogalpaulina:
Custom error message on Student Orientation widget

Screen Shot 2020-10-29 at 2 21 12 PM

----

PR considerations checklist:

### DIFF
--- a/components/portal/widgets/controllers.js
+++ b/components/portal/widgets/controllers.js
@@ -690,7 +690,7 @@ define(['angular', 'moment'], function(angular, moment) {
       // Custom error message on Student Orientation widget
       var studentOrientationWidgetError = function() {
       return {
-        title: student-orientation,
+        title: 'student-orientation',
         mdIcon: 'help',
         widgetType: 'generic',
         widgetConfig: {
--- a/components/portal/widgets/controllers.js
+++ b/components/portal/widgets/controllers.js
@@ -671,6 +671,8 @@ define(['angular', 'moment'], function(angular, moment) {
               eval($scope.widget.widgetConfig.evalString)) {
               $scope.isEmpty = true;
             }
+          } else if (($scope.widget === 'student-orientation') && !data) {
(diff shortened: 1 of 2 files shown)


components/portal/widgets/controllers.js
```diff
@@ -671,6 +671,8 @@ define(['angular', 'moment'], function(angular, moment) {
               eval($scope.widget.widgetConfig.evalString)) {
               $scope.isEmpty = true;
             }
+          } else if (($scope.widget === 'student-orientation') && !data) {
```
apetro, line 674: Clever. I'm fine with this as a one-off. 

Nonetheless, there are tradeoffs here.

This is coding into the framework configuration specific to a single app directory entry. If we need this feature for more than one entry, it might be worth it to refactor it into the framework as a shared feature, as in error configuration or an error template or so goes into the data model for app directory entries and any entry that declares a custom error template can use this feature.

So. :shipit: . But when it becomes annoying to have to cut a framework release to edit this error message or when we do to add another widget id to this conditional, let's refactor. :)

apetro: Aside: this is making me think about, what if we embraced this approach, and merged all the app directory data into the framework. Stop trying to store that in and read it from a database. Just code. Might be worth it for easier testing and reducing runtime dependency.

apetro (approved):

components/portal/widgets/controllers.js
nogalpaulina: Thank you for reviewing Andrew! 
I've been thinking about this solution, and I agree with you that there are more efficient ways to accomplish this. I'm going to close this MR, and move forward with [handling the error message in the entity file](https://git.doit.wisc.edu/myuw-overlay/entities/-/merge_requests/2726).

nogalpaulina:
Moving forward with another solution.

PR #1030, opened 2021-11-11
Reset new layout

nogalpaulina:
+ Add feature flag to check which layout is used and utilize new boolean values to conditionally display Reset layout option.
+ Create new newLayoutResetService service to handle layout reset.

PR considerations checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ It is those war files that are being versioned.
 
 ## Next
 
+## 21.0.3 - 2021-11-11
++ Add feature flag to check which layout is used and utilize new boolean values to conditionally display Reset layout option.
++ Create new newLayoutResetService service to handle layout reset.
+
+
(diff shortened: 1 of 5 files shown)


apetro:
The new layout backend expects a POST and provides feedback on whether it succeeded via the HTTP status code. So this UI needs to POST to that URL on button click, and then ideally reflect the status response it gets back.

I (or you) *could* implement a UI in the `/layout-service/` itself for layout reset, more like the way the old layout reset works hyperlinking to an application to provide this experience, but... seems better to not bother the user with the extra clicks and navigation and keeps the `/layout-service/` more tightly an API microservice than an end-user-facing application.

So. I think we should go the user-does-not-navigate-away-from-settings-page approach where the button click causes an async POST, but, I'm open to negotiation on how we implement this.

apetro (approved):
Let's try it.

